Issue #69, opened 2026-04-13
Phase 1: MCP Auth — Config model & validation


usepowershell:
## Overview
Create the configuration model and validation infrastructure for MCP authentication. This is the foundation all other auth phases depend on.

## Tasks
- Create \AuthenticationConfiguration\ POCO classes binding to \ppsettings.json\:
  - \AuthenticationConfiguration\ (root)
  - \AuthorizationPolicyConfiguration\ (default policy)
  - \AuthSchemeConfiguration\ (per-scheme: JwtBearer or ApiKey)
  - \ClaimsMappingConfiguration\
  - \ApiKeyDefinition\
  - \ProtectedResourceConfiguration\ (RFC 9728 fields)
- Extend \FunctionOverride\ in \PowerShellConfiguration.cs\ with:
  - \RequiredScopes\ (\List<string>\)
  - \RequiredRoles\ (\List<string>\)
  - \AllowAnonymous\ (\ool\, default \alse\)
- Implement \IValidateOptions<AuthenticationConfiguration>\ validator:
  - If \Enabled = true\, at least one scheme must be defined
  - JWT Bearer schemes must have \Authority\ set
  - API Key schemes must have at least one key defined
- Extend \ConfigurationHealthCheck\ with auth status fields
- Add \Authentication\ section to \ppsettings.json\ with \Enabled: false\ default

## Files
- **New:** \PoshMcp.Server/Authentication/AuthenticationConfiguration.cs\
- **New:** \PoshMcp.Server/Authentication/AuthenticationConfigurationValidator.cs\
- **Modified:** \PoshMcp.Server/PowerShell/PowerShellConfiguration.cs\
- **Modified:** \PoshMcp.Server/Health/ConfigurationHealthCheck.cs\
- **Modified:** \PoshMcp.Server/appsettings.json\

## Notes
All properties should have sensible defaults so \"Authentication": { "Enabled": true }\ works with zero extra config. Auth is **disabled by default** to preserve backward compatibility.

## Dependencies
None — this is the first phase.

usepowershell:
Phase 1 implementation is complete. PR: https://github.com/usepowershell/PoshMcp/pull/77

PR #79, opened 2026-04-13
test: auth configuration unit tests (Phase 1)

### LINKED ISSUE #75: Phase 7: MCP Auth — Test suite (unit + integration + functional)
## Overview
Comprehensive test coverage for the MCP authentication implementation across unit, integration, and functional test layers.

### \AuthenticationConfigurationTests.cs\
- Valid minimal config (\Enabled: true\ with one JWT scheme) validates successfully
- Missing authority on JWT scheme fails validation
- \Enabled: true\ with no schemes fails validation
- Default values applied correctly (\Enabled: false\, etc.)
- Per-tool override resolution: function override > default policy > disabled

### \ToolAuthorizationFilterTests.cs\
- Authenticated user with correct scopes → allowed
- Authenticated user with missing scopes → 403-style MCP error
- Unauthenticated user, tool requires auth → 401-style MCP error
- Unauthenticated user, tool has \AllowAnonymous: true\ → allowed
- Per-tool \RequiredScopes\ override takes precedence over \DefaultPolicy\
- Auth disabled globally → filter is no-op

### \ApiKeyAuthenticationHandlerTests.cs\
- Valid API key in default header → authenticated with correct claims
- Invalid API key → \AuthenticateResult.Fail\
- Missing header → no result (fall through to next scheme)
- Custom header name from config works
- Key with scopes/roles → correct claims on principal

### \AuthenticationIntegrationTests.cs\ (HTTP transport)
- HTTP request with valid Bearer token → 200 + tool result
- HTTP request with no token, auth enabled → 401 response
- HTTP request with token, insufficient scopes for tool → 403-style MCP error
- Health endpoint (\/health\) accessible without auth token
- Protected Resource Metadata endpoint (\/.well-known/oauth-protected-resource\) returns correct JSON
- Auth disabled (\Enabled: false\) → all tools accessible (backward compat preserved)

### \PerToolAuthorizationTests.cs\
- Tool A has \AllowAnonymous: true\, Tool B requires auth → A works unauthenticated, B returns error
- Tool A requires scope \	ools:read\, Tool B requires scope \	ools:write\ → user with only \	ools:read\ can call A but not B
- \	ools/list\ for partial-access user returns only authorized tools
- Default policy applies to tools without \FunctionOverrides\ entries

## Files
- **New:** \PoshMcp.Tests/Unit/AuthenticationConfigurationTests.cs\
- **New:** \PoshMcp.Tests/Unit/ToolAuthorizationFilterTests.cs\
- **New:** \PoshMcp.Tests/Unit/ApiKeyAuthenticationHandlerTests.cs\
- **New:** \PoshMcp.Tests/Integration/AuthenticationIntegrationTests.cs\
- **New:** \PoshMcp.Tests/Functional/PerToolAuthorizationTests.cs\

## Dependencies
Depends on #69 (Phase 1: Config model) — can be written in parallel with implementation phases

usepowershell:
Unit tests for #69 Phase 1 auth config. Depends on #77.

Closes #75 (partial — more tests will be added as phases complete).

### DIFF
--- a/PoshMcp.Tests/Unit/AuthenticationConfigurationValidatorTests.cs
+++ b/PoshMcp.Tests/Unit/AuthenticationConfigurationValidatorTests.cs
@@ -1,0 +1,197 @@
+using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+using PoshMcp.Server.Authentication;
+using Xunit;
+
+namespace PoshMcp.Tests.Unit;
+
+public class AuthenticationConfigurationValidatorTests
+{
+    private readonly AuthenticationConfigurationValidator _validator = new();
+
+    [Fact]
+    public void Validate_WhenDisabled_ReturnsSuccess()
+    {
+        var config = new AuthenticationConfiguration
+        {
+            Enabled = false,
+            Schemes = new Dictionary<string, AuthSchemeConfiguration>()
+        };
+
+        var result = _validator.Validate(null, config);
+
+        Assert.Equal(ValidateOptionsResult.Success, result);
+    }
+
+    [Fact]
+    public void Validate_WhenEnabled_NoSchemes_ReturnsFail()
+    {
+        var config = new AuthenticationConfiguration
+        {
+            Enabled = true,
+            Schemes = new Dictionary<string, AuthSchemeConfiguration>()
+        };
+
+        var result = _validator.Validate(null, config);
+
+        Assert.True(result.Failed);
+        Assert.Contains(result.Failures, f => f.Contains("no schemes"));
+    }
+
+    [Fact]
+    public void Validate_JwtBearer_MissingAuthority_ReturnsFail()
+    {
+        var config = new AuthenticationConfiguration
+        {
+            Enabled = true,
+            Schemes = new Dictionary<string, AuthSchemeConfiguration>
+            {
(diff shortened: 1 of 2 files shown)


usepowershell (commented):
## Farnsworth Review — PR #79: Auth Configuration Unit Tests (Phase 1)

**Verdict: APPROVE** ✅

### Correctness
Tests cover the right behavioral boundaries: disabled-returns-success, enabled-without-schemes fails, JWT missing authority, API key empty keys, invalid protected resource URI, multiple error accumulation, and FunctionOverride default/serialization. All assertion patterns are correct.

### Coverage
Good breadth — covers the validator and the FunctionOverride auth property additions. The JSON deserialization test for FunctionOverride is a smart inclusion; it catches config binding issues that unit tests on the class alone would miss.

### Minor observations
- The `FunctionOverrideAuthPropertiesTests` class tests property getters/setters which is low-signal on its own, but the `DeserializesFromJson` test makes the file worthwhile.
- Tests correctly reference types already on `main` — no compilation issues.

### Suggestion (non-blocking)
Consider adding a test for unknown scheme types (e.g., `Type = 'OAuth2'`) — the validator currently silently accepts them. This can be deferred to a follow-up.

Clean work, Fry.

PR #84, opened 2026-04-13
fix: handle warning stream content during OOP server startup

### LINKED ISSUE #78: Error starting out of process server on warning stream content.
Failed to parse OOP stdout JSON: WARNING: Current tenant context is set to d91aa5af-8c1e-442c-b77c-0b92988b387b. Use Set-TenantContext to change the tenant context.
      System.Text.Json.JsonReaderException: 'W' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
         at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
         at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
         at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
         at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
         at System.Text.Json.Utf8JsonReader.Read()
         at System.Text.Json.JsonDocument.Parse(ReadOnlySpan`1 utf8JsonSpan, JsonReaderOptions readerOptions, MetadataDb& database, StackRowStack& stack)
         at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 utf8Json, JsonReaderOptions readerOptions, Byte[] extraRentedArrayPoolBytes, PooledByteBufferWriter extraPooledByteBufferWriter, Boolean allowDuplicateProperties)
         at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 json, JsonDocumentOptions options)
         at PoshMcp.Server.PowerShell.OutOfProcess.OutOfProcessCommandExecutor.ReadLoopAsync() in C:\Users\stmuraws\source\poshmcp\PoshMcp.Server\PowerShell\OutOfProcess\OutOfProcessCommandExecutor.cs:line 429

usepowershell:
## Summary

Fixes the error that occurs when starting the out-of-process server when there is content on the warning stream.

## Root Cause

PowerShell modules (e.g., Az) emit Write-Warning during Import-Module, which in the OOP subprocess context leaks to stdout and corrupts the ndjson protocol channel. The C# ReadLoopAsync then tries to parse the WARNING: line as JSON, fails, and logs an alarming Warning-level entry with a full JsonReaderException stack trace.

### oop-host.ps1 (primary fix)
- Added -WarningAction SilentlyContinue -WarningVariable to all Import-Module and Install-Module calls
- Captured warnings are forwarded to stderr via Write-Diag, preserving observability without polluting stdout

### OutOfProcessCommandExecutor.cs (defensive fix)
- Added IsNonJsonPowerShellStreamLine() helper detecting WARNING:, VERBOSE:, DEBUG:, INFORMATION:, ERROR: prefixes
- Such lines are skipped at Debug level with no exception thrown
- Demoted catch(JsonException) from LogWarning(ex,...) to LogDebug to eliminate alarm fatigue

## Testing

1. Configure PoshMcp to import an Az module that emits tenant-context warnings on import
2. Start the OOP server - previously shows 'Failed to parse OOP stdout JSON: WARNING:...' with stack trace
3. After fix: warnings routed to stderr/diagnostics, server starts cleanly

Fixes #78

### DIFF
--- a/.squad/agents/farnsworth/history.md
+++ b/.squad/agents/farnsworth/history.md
@@ -113,3 +113,9 @@ Current Priorities:
 - #86: Enhancement — add `--use-default-display-properties` global flag (consistency with other Performance bool flags)
 - #87: Tech debt — warn when `--set-auth-enabled true` used with empty `Authentication.Schemes`
 - #88: Tests — add unit tests for all 4 new flags and `settingsChanged` counter in `ProgramCliConfigCommandsTests`
+### 2026-07-15: PR #83 re-review (auth metrics Phase 6)
+
+**Verdict:** APPROVED
+**Issue resolved:** McpMetrics dual-instance bug fixed by Bender. Auth filters now registered as DI singletons with factory lambdas resolving `McpMetrics` via `sp.GetRequiredService<McpMetrics>()`. No manual `new McpMetrics()` construction in auth path. Deferred capture pattern for filter variables (assigned post-`app.Build()`) is safe — lambdas execute only at request time.
+**Non-blocking nit:** Redundant LINQ lookup in `ApiKeyAuthenticationHandler` (`Options.Keys.FirstOrDefault(k => k.Key == apiKey).Key` after `TryGetValue` already succeeded).
+**Build:** 0 errors on branch.
--- a/PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessCommandExecutor.cs
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessCommandExecutor.cs
@@ -424,6 +424,15 @@ public class OutOfProcessCommandExecutor : ICommandExecutor
 
                 _logger.LogDebug("OOP stdout: {Line}", line);
 
+                // PowerShell warning/info/verbose/debug streams can leak to stdout from
+                // modules that call Write-Warning (or similar) directly. These lines are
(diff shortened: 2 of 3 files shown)


usepowershell:
This is one of the tradeoffs with out of process.  Either we use a higher fidelity transport (like CLIXML, which maintains the streams concept), create a runspace wrapper and programmatically interact with it in the separate process, or do what we are doing here and parse text.

usepowershell (commented):
Agreed — and this PR deliberately leans into the text-parsing tradeoff while minimising its downsides.

The two alternatives are worth keeping on the radar:

- **CLIXML** would give us proper stream separation but adds serialisation overhead and tighter coupling to the PowerShell host format (which varies across versions).
- **Runspace wrapper** (in-process with a separate AppDomain/process boundary) is more robust but significantly increases implementation complexity.

For the current use-case the text-parsing approach is pragmatic: module warnings are rare, IsNonJsonPowerShellStreamLine is a cheap fast-path check, and the primary fix lives in oop-host.ps1 (suppressing warnings at the source). If we ever need true stream fidelity — for instance to surface structured warning objects as MCP annotations — CLIXML would be the natural upgrade path. Happy to open a tracking issue for that if useful.

PR #93, opened 2026-04-14
Warn when --set-auth-enabled true is used with no Authentication.Schemes configured

### LINKED ISSUE #87: Tech debt: warn when --set-auth-enabled true is used with no Authentication.Schemes configured
## Context

Observed during review of PR #85 (extend update-config to support all configuration settings).

## Observation

When a user runs `poshmcp update-config --set-auth-enabled true`, the command succeeds and writes the config without checking whether `Authentication.Schemes` is populated. If no schemes are configured, the resulting config will fail `AuthenticationConfigurationValidator` at server startup — but the user gets no warning at the time of the change.

This is consistent with the existing pattern (`validate-config` exists for this purpose and is documented), but a proactive warning at the CLI level would improve the UX and catch a common misconfiguration earlier.

## Suggested Work

In `UpdateConfigurationFileAsync`, after writing `Authentication.Enabled = true`, check whether `root["Authentication"]?["Schemes"]` is null or an empty array. If so, emit a warning line to stderr (or stdout with a "WARNING:" prefix) along the lines of:

`WARNING: Authentication.Enabled set to true but Authentication.Schemes is empty. Run 'poshmcp validate-config' to verify your configuration.`

This should not block the write — it is advisory only.

usepowershell:
## Summary

When a user runs \poshmcp update-config --set-auth-enabled true\ without any \Authentication.Schemes\ configured, the server will fail \AuthenticationConfigurationValidator\ at startup — but the user gets no warning at the time of the change.

## Changes

In \UpdateConfigurationFileAsync\, after writing \Authentication.Enabled = true\, added a check: if \Authentication.Schemes\ is null or an empty array, emit a warning to stderr:

\\\
WARNING: Authentication.Enabled set to true but Authentication.Schemes is empty. Run 'poshmcp validate-config' to verify your configuration.
\\\

The check is **advisory only** — the write proceeds unconditionally.

Fixes #87

### DIFF
--- a/PoshMcp.Server/Program.cs
+++ b/PoshMcp.Server/Program.cs
@@ -1653,6 +1653,15 @@ public class Program
             var authentication = GetOrCreateObject(root, "Authentication");
             authentication["Enabled"] = request.SetAuthEnabled.Value;
             boolUpdateApplied++;
+
+            if (request.SetAuthEnabled.Value)
+            {
+                var schemes = authentication["Schemes"]?.AsArray();
+                if (schemes == null || schemes.Count == 0)
+                {
+                    Console.Error.WriteLine("WARNING: Authentication.Enabled set to true but Authentication.Schemes is empty. Run 'poshmcp validate-config' to verify your configuration.");
+                }
+            }
         }
 
         var advancedPromptedFunctionCount = 0;


usepowershell:
## ✅ APPROVED — Farnsworth (Lead/Architect)

Advisory warning is correctly scoped: only fires when `SetAuthEnabled.Value` is true, uses `authentication['Schemes']?.AsArray()` (equivalent to `root['Authentication']['Schemes']`), null-or-empty guard is correct, `Console.Error.WriteLine` writes to stderr, and the write proceeds unconditionally. Warning message includes actionable guidance pointing to `validate-config`.

**Well done:** Minimal, focused change — advisory only with no side effects on the write path.

PR #95, opened 2026-04-14
feat: skip unserializable parameter types in MCP tool schema generation

### LINKED ISSUE #89: Dealing with unserializable parameter types
When there is a non-mandatory parameter that has a data type which can't be serialized to JSON effectively, the tool should skip adding that parameter.

If the parameter is mandatory in a parameter set but optional or not required in others, just skip that parameter set.

If there are no parameter sets where the parameter is not mandatory, then skip the command.

usepowershell:
## Summary

Fixes #89

When a PowerShell command has parameters whose types cannot be meaningfully serialized to JSON, PoshMcp now handles them gracefully instead of exposing broken or misleading schema entries.

## Behavior

| Scenario | Action |
|---|---|
| Optional parameter with unserializable type | Drop that parameter from the schema silently |
| Mandatory parameter with unserializable type in a parameter set | Skip the entire parameter set |
| All parameter sets skipped for a command | Skip the command (no MCP tool exposed); warning logged |

## Unserializable Types

PowerShellParameterUtils.IsUnserializableType() returns 	rue for:
- PSObject — opaque wrapper with no fixed schema
- ScriptBlock — executable code, not data
- System.Object — too generic for a meaningful schema
- Pointer types (IntPtr, UIntPtr, T*, T&)
- Delegate-derived types (callbacks, Action, Func<>, …)
- Stream-derived types (binary channels)
- WaitHandle-derived types (OS sync primitives)
- System.Reflection.Assembly — reflection handle
- System.Management.Automation.PowerShell — PS automation instance
- All System.Management.Automation.Runspaces.* types (Runspace, RunspacePool, …)
- Arrays whose element type is unserializable

## Tests

Added UnserializableTypeTests.cs with 33 unit tests covering both the positive (serializable) and negative (unserializable) categories.

### DIFF
--- a/PoshMcp.Server/PowerShell/PowerShellAssemblyGenerator.cs
+++ b/PoshMcp.Server/PowerShell/PowerShellAssemblyGenerator.cs
@@ -207,16 +207,24 @@ public class PowerShellAssemblyGenerator
                 var commandList = commands.ToList();
                 foreach (var command in commandList)
                 {
+                    var anyParameterSetGenerated = false;
                     foreach (var parameterSet in command.ParameterSets)
                     {
                         try
                         {
-                            GenerateMethodForCommand(typeBuilder, command, loggerField, runspaceField, logger, parameterSet);
+                            var generated = GenerateMethodForCommand(typeBuilder, command, loggerField, runspaceField, logger, parameterSet);
+                            if (generated)
+                                anyParameterSetGenerated = true;
                         }
                         catch (Exception ex)
                         {
                             logger.LogError(ex, $"Failed to generate method for command {command.Name}: {ex.Message}");
                         }
+                    }
+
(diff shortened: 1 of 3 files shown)


usepowershell:
## ✅ APPROVED — Farnsworth (Lead/Architect)

Solid implementation. `IsUnserializableType` covers the correct categories: pointer/by-ref, IntPtr/UIntPtr, PSObject, ScriptBlock, System.Object, delegates, streams, WaitHandle, Assembly, PowerShell, Runspaces namespace, and recursive array elements. Nullable unwrapping correct. Three-tier handling (mandatory → skip set, optional → drop param, all skipped → skip command) follows spec exactly. 33 unit tests provide good positive/negative coverage.

**Well done:** Proper separation — type classification in utils, integration in generator, tests alongside.

Issue #65, opened 2026-04-11
OOP: Experiment with runspace pool parallelism vs multiple processes


usepowershell:
## Summary

Experiment with using a PowerShell runspace pool inside the single `pwsh` subprocess (via `oop-host.ps1`) for parallel command execution, and compare performance and behavior against spawning multiple separate `pwsh` processes.

**Spec reference:** `specs/out-of-process-execution.md` — Open Question #1

**Context:** Steven's direction: *"We should experiment and look at creating a runspace pool in the one process and using new runspaces for each instance needed. Then we should compare performance and behavior to separate processes."*

### Option A: Runspace Pool in oop-host.ps1
- Modify `oop-host.ps1` to create a `[runspacefactory]::CreateRunspacePool()` on startup
- Route `invoke` requests to available runspaces from the pool
- Handle concurrent ndjson responses (correlate by request id)
- Measure: throughput, memory, isolation guarantees

### Option B: Multiple pwsh Processes  
- Modify `OutOfProcessCommandExecutor.cs` to manage N subprocess instances
- Round-robin or queue-based dispatch
- Measure: throughput, memory, startup cost, isolation guarantees

### Comparison Criteria
- Throughput (concurrent invocations per second)
- Memory overhead per concurrent slot
- Isolation (does a crash in one invocation affect others?)
- Startup latency
- Complexity of implementation

## Acceptance Criteria
- [ ] Prototype runspace pool approach in oop-host.ps1
- [ ] Prototype multi-process approach in OutOfProcessCommandExecutor
- [ ] Benchmark comparison with documented results
- [ ] Recommendation for which approach to adopt

## Dependencies
- Phases 1-4 complete (subprocess lifecycle, discovery, invocation working)
- Best attempted after Phase 6 integration tests validate the single-process baseline

usepowershell:
Spec 004 milestone created and follow-up issues filed.

**Milestone:** [Spec 004 - Out-of-Process PowerShell Execution](https://github.com/usepowershell/PoshMcp/milestone/5)

**Follow-up issues (in dependency order):**

| # | Title | Plan ref | Blocked by |
|---|-------|----------|------------|
| #189 | OOP: Bug-fix — clear `System.Management.Automation.ParseException: At line:20 char:37
+ **Operational note:** EMU policy on `usepowershell/PoshMcp` blocks bo …
+                                     ~~
[...]

PR #187, opened 2026-05-06
docs(004): runspace pool vs multi-process experiment plan (Refs #65)

usepowershell:
**🧪 Hermes (PowerShell Expert)**

R&D plan for Issue #65 — comparing in-process runspace pooling against a pool of pwsh subprocesses for OOP execution.

## Summary

This is a **plan-only** PR. No code changes. The issue stays open through the prototype work that follows.

The plan covers:

1. **Current state audit** — Phases 1–4 of spec 004 are complete and shipping. Single-subprocess executor serializes invokes through `_sendLock`; the ndjson protocol is already async-correlated by id, so the protocol layer is parallel-ready — only the host is serialized.
2. **Option A — Runspace pool inside one subprocess** — `[runspacefactory]::CreateRunspacePool` with InitialSessionState pre-warmed with modules. Concrete handling for the three concurrency hazards: synchronized stdout writer, per-runspace `$Error` clearing, and stream-pollution mitigations. Shipped as a new `oop-host-pool.ps1` selected by a `SubprocessHostMode` config flag so the existing host stays as a known-good baseline.
3. **Option B — Pool of N subprocesses** — `OutOfProcessSubprocessPool` over a `Channel<OutOfProcessHost>` for FIFO fairness; per-host crash/restart unchanged; `SubprocessPoolSize: 1` collapses to today's behavior.
4. **Benchmark harness** — separate `PoshMcp.Benchmarks` console project, BenchmarkDotNet for latency/throughput plus a custom harness for crash/recovery tests. Eight scenarios (CPU-light, CPU-bound, I/O-bound, network-shaped, heavy serialization, cold start, crash recovery, isolation) with explicit pass/fail criteria including an isolation gate.
5. **Recommended phasing** — six follow-up issues, starting with extracting `OutOfProcessHost` so per-process state is reusable. Prototypes (2/3) and harness (4) parallelize after that refactor.

## Key trade-off the plan flags

Option A shares an AppDomain. If a module like Az corrupts a type accelerator, the corruption survives runspace disposal — meaning we get parallelism but lose the strongest isolation guarantee that motivated OOP in the first place. The benchmark harness measures this directly and the plan recommends defaulting to B if neither cleanly passes the isolation gate.

## Open questions for review

Documented in §6 of the plan — default `N` for the pool, whether to ship Option A at all if B wins, fidelity of the network-shaped scenario, and whether cancellation propagation should be in-scope or a separate issue.

## File touched

- `specs/004-out-of-process-execution/runspace-pool-experiment-plan.md` (new)

Refs #65


usepowershell:
## Verification Report — runspace-pool-experiment-plan.md

**Fact-checker:** Cubert
**Date:** 2026-05-06
**Method:** Cross-checked every API name, file path, type signature, and behavioral claim against the source tree on `squad/65-runspace-pool-experiment-plan` (and against decisions.md / Issue #65). This is a fact-check, not an architectural verdict.

### Claims Reviewed

#### ✅ Verified
- **Subprocess launch line** — `OutOfProcessCommandExecutor` launches `pwsh -NoProfile -NonInteractive -File oop-host.ps1`, attaches stdin/stdout/stderr, hooks `Process.Exited`. Confirmed in `PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessCommandExecutor.cs:74,86`.
- **Disposal pattern** — `shutdown` request → wait → `Kill(entireProcessTree: true)` fallback. Confirmed at `OutOfProcessCommandExecutor.cs:316`.
- **`_sendLock`** is `SemaphoreSlim(1, 1)`. Confirmed at `OutOfProcessCommandExecutor.cs:29`.
- **`_pending`** is `ConcurrentDictionary<string, TaskCompletionSource<JsonElement>>`. Confirmed at `OutOfProcessCommandExecutor.cs:30`.
- **GUID `id`** — `Guid.NewGuid().ToString("N")`. Confirmed at `OutOfProcessCommandExecutor.cs:375`.
- **Background `ReadLoopAsync`** — Confirmed at `OutOfProcessCommandExecutor.cs:440`.
- **`IsNonJsonPowerShellStreamLine` backstop** — Confirmed at `OutOfProcessCommandExecutor.cs:558`.
- **ndjson methods** `ping` / `setup` / `discover` / `invoke` / `shutdown` — all four handler functions exist in `oop-host.ps1` (`Invoke-PingHandler`, `Invoke-ShutdownHandler`, `Invoke-SetupHandler`, `Invoke-DiscoverHandler`, `Invoke-InvokeHandler`).
- **Discover walks `Get-Command` + `ParameterSets`** — Confirmed in `oop-host.ps1:385,469`.
- **Invoke uses `ConvertTo-Json -Depth 4 -Compress`** — Confirmed at `oop-host.ps1:572`. (Note: `Send-Response` uses `-Depth 10` for the wrapper envelope; the user-payload depth is 4 as the plan states.)
- **`hadErrors` derived from `$Error.Count`** — Confirmed at `oop-host.ps1:561-565`.
- **Setup mirrors `PowerShellEnvironmentSetup` ordering** — Comment at `oop-host.ps1:82` documents this intent; setup parameters (`installModules`, `trustPSGallery`, `startupScriptPath`, `startupScript`) line up with the .NET environment-setup surface.
- **`RemoteToolSchema`** type exists at `PoshMcp.Server/PowerShell/OutOfProcess/RemoteToolSchema.cs:9`.
- **External APIs cited** — `[runspacefactory]::CreateRunspacePool(min, max, $iss, $Host)`, `InitialSessionState`, `[powershell]::Create()`, `BeginInvoke`, `AsyncCallback`, `PSEventSubscriber`, `ReaderWriterLockSlim`, `System.Threading.Channels.Channel<T>`, `Environment.ProcessorCount`, `BenchmarkDotNet`, `HttpListener` — all real, all spelled correctly.
- **`OutOfProcessHost` does not yet exist** — Confirmed (no such file under `PoshMcp.Server/PowerShell/OutOfProcess/`). The plan correctly lists this as Phase 1 of the follow-up phasing, not as an existing artifact.
- **No `PoshMcp.Benchmarks` project exists** — Confirmed (no matching `.csproj`). Consistent with the plan proposing to create it.
- **Issue #65** is open and assigned to Hermes via `squad:hermes` label. Confirmed.

#### ⚠️ Needs revision
- **"Phases 1–4 complete"** (§1 heading and PR description). The current spec at `specs/004-out-of-process-execution/spec.md` is marked `Status: Draft` and contains no phase manifest — there is no `plan.md` or `tasks.md` in that spec directory enumerating Phases 1–4. The phrase originates from Issue #65's own "Dependencies" note. The substantive claim (subprocess lifecycle, ndjson protocol, discovery, invoke, setup, shutdown all shipping) is **true** and reproducible in the source. Recommend either (a) rephrasing to drop the "Phases 1–4" label and describe the shipped capabilities directly, or (b) landing a phase manifest in spec 004 first so the label is anchored.
- **Path drift in upstream issue** — Issue #65 references `specs/out-of-process-execution.md` (no `004-` prefix, no directory). The plan correctly uses `specs/004-out-of-process-execution/`. The plan is right; the issue is stale. Worth a brief footnote so future readers don't chase the wrong path.
- **`$Error` clearing is a pre-existing bug, not just a pool-mode hazard** (§2, "ndjson concurrency hazards" point 3). Verified at `oop-host.ps1:558-565`: the current single-runspace `Invoke-InvokeHandler` does **not** clear `$Error` before `& $commandName`, so `hadErrors` already reflects errors from prior invocations leaking across calls in the existing code. The plan flags this as a pool-mode concern but does not note it is **already** wrong in single-runspace mode. Consider promoting to a separate bug-fix issue independent of the experiment outcome.

#### 🔍 Needs investigation
- **Baseline capture for the pass/fail criteria** (§4). The criteria reference "the single-subprocess baseline" but the plan doesn't specify whether the baseline is measured by the same harness in the same run as A/B, or pre-captured. If the harness establishes the baseline simultaneously, the criteria are well-defined; if it relies on a prior baseline run, that run is not yet specified anywhere. Worth one sentence in §4 to pin this down.
- **`HttpListener` portability** (§4 network-shaped scenario). `System.Net.HttpListener` is supported on .NET 10 across platforms, but on Windows it can require URL ACL setup for non-loopback bindings. Using `127.0.0.1` with an ephemeral port (`:0`) avoids this — recommend the plan say so explicitly to head off a portability footgun in the harness.

### Counter-Hypotheses

- **"The protocol layer does not require serialization."** Counter-checked: `_sendLock` exists to prevent interleaved `WriteAsync` on `_process.StandardInput` from the .NET side, not the host side. The plan's §2 "Files touched" explicitly preserves `_sendLock`, so the claim and the recommended action are consistent. ✅
- **"Type-accelerator corruption survives runspace disposal."** True at the AppDomain level for static state, but most Az cmdlet state is per-runspace. The §4 isolation scenario is the right place to falsify or confirm this on a per-module basis — plan already specifies it. ✅
- **"`SubprocessPoolSize: 1` collapses to current behavior."** Verified by inspection: a `Channel<OutOfProcessHost>` with one entry yields strict serial dispatch through the same single-process executor path. ✅

### Cross-Check Against `.squad/decisions.md`

- Decisions ledger entry at `.squad/decisions.md:799` (Hermes, 2026-05-06) accurately summarizes the plan: two prototypes, BenchmarkDotNet harness, six follow-up issues, isolation as a pass/fail gate. No conflicts with prior decisions.

### Recommendation

**Proceed.** The plan's technical claims are accurate, file paths and API references check out, and the proposed phasing is internally consistent. Two items worth addressing before or alongside merge:

1. Soften or anchor the "Phases 1–4 complete" label (no phase manifest exists in spec 004 today).
2. Note that the existing single-runspace `$Error` non-clearing is a pre-existing bug worth its own fix issue, independent of the pool experiment outcome.

Neither blocks landing the plan.

usepowershell:
## Architectural review — runspace pool vs multi-process experiment plan

Overall: this is a strong plan. The framing is honest about the tradeoff (parallelism vs blast-radius isolation), the two prototype designs are concrete enough to build from, and the phasing is mostly right. Approving the direction; raising several substantive concerns to fold in before or during the prototype issues.

### Alignment with spec 004

The plan respects the existing acceptance criteria: subprocess restart on crash (FR-049), per-request timeout + kill/restart (FR-050), serialized ndjson framing (FR-051), graceful shutdown (FR-053). Option B preserves the current isolation model 1:1 across N hosts, which keeps US1/US2 honest. Option A weakens the US2 isolation contract by design — worth calling out explicitly in the spec if Option A wins, because "subprocess isolation protects the server" is no longer strictly true at the runspace level.

### Option A concerns

1. **`[Console]::Out` redirection per runspace is not actually a thing.** §2 "Stream pollution" suggests "replace `[Console]::Out` for the pool runspaces with a `TextWriter.Null` adapter." `Console.Out` is a process-global static — you cannot scope it per runspace via ISS. The realistic mitigations are (a) `$Host.UI` overrides via a custom `PSHost` implementation passed to `CreateRunspacePool`, and (b) the existing `IsNonJsonPowerShellStreamLine` backstop. Worth correcting before prototype kickoff so we don't burn a day discovering this.

2. **Setup-while-running race is underspecified.** §2 "Setup model" says "Setup is rare and cannot race with invokes." On the host side that's fine if the .NET executor serializes setup against in-flight invokes — but `OutOfProcessCommandExecutor.SendRequestAsync` does not gate setup against in-flight invokes today (only stdin writes via `_sendLock`). Closing+reopening the pool while invokes are queued/active will either drop them or hang. Plan should specify the quiesce protocol: drain pending → close pool → rebuild ISS → open pool → resume.

3. **Per-runspace `$Error` handling is correct but partial.** Clearing `$Error` before each invoke addresses one half. Streams (warning, verbose, information) need the same per-pipeline scoping or `hadErrors` semantics will diverge from the single-runspace baseline. Worth a one-liner.

4. **Backpressure observability.** `SubprocessQueueLimit` is fine as a knob, but without a metric (queue depth, lease wait time) operators will have no signal to tune it. Add a counter to the prototype — minor scope, large operability win.

### Option B concerns

5. **Setup fail-fast vs degraded-pool.** §3 "Lifecycle" picks fail-fast. Reasonable for parity with current behavior, but if the failure is transient (PSGallery hiccup), losing the entire server on a 1-of-N setup failure is a regression vs the single-host baseline (which would just fail one start and the next request retries). Recommend: fail-fast on the *first* host's setup, then per-host retries with backoff for the remainder. The first host's setup is the meaningful smoke test.

6. **Channel<OutOfProcessHost> + Process.Exited race.** When a host crashes mid-lease (before being returned to the channel), the dispatch loop has no way to know the leased host is dead until the consumer's `await` on the response throws. Need to specify: on `Process.Exited` for a leased host, the replacement spawn must **not** be added to the channel until ping+setup complete *and* the leased slot must be reconciled (i.e., we don't end up with a pool of N+1 or N-1 silently). A `ConcurrentDictionary<Process, HostState>` alongside the channel is probably needed.

7. **Discovery once vs per-host.** §3 says "discovery happens once, against any one subprocess." Correct in steady state, but if hosts are restarted independently, the new host has the modules imported (via setup) but the schema cache lives on the .NET side keyed off the pool, not the host. This is fine — just worth a comment in the plan so the next reader doesn't assume per-host discovery is needed.

### Benchmark harness

8. **Crash recovery test for Option A is not measuring what we think it is.** §4 "Crash recovery" induces process exit. In Option A there's only one process, so this measures the same thing as the baseline (process restart). To actually test Option A's isolation story, induce **runspace-level corruption** (e.g., pollute a type accelerator, mutate a shared PowerShell static) in one invoke and verify a parallel invoke is unaffected. That's the real isolation gate for A. Process-kill recovery is the gate for B.

9. **Pass/fail criterion "≥ 4× baseline at 10 concurrent clients" is too aggressive for CPU-bound.** A is constrained by ThreadPool + GC contention inside one process; B is constrained by IPC + per-process memory. Neither will hit 4× on the CPU-bound scenario at any reasonable N. Recommend per-scenario thresholds: aggressive (≥ 4×) on I/O-bound and network-shaped, modest (≥ 1.5×) on CPU-bound, parity (within 20%) on CPU-light. Otherwise we'll reject a winning design on a scenario that isn't its job.

10. **Output normalization scenario should also vary payload size.** `Get-Process | Select-Object -First 50` is a fixed shape. Add a parameterized variant (`-First 10/100/1000`) — this is where Option B's IPC overhead gets exposed and where Option A's shared `ConvertTo-Json` may contend.

11. **Network-shaped fidelity** (your open question §6.3): local `HttpListener` is sufficient for the prototype gate. Real-Azure validation belongs in a follow-up acceptance test, not the bench harness. Don't block the comparison on tenant access.

### Phasing

12. **Issue #1 (Extract OutOfProcessHost) needs explicit test coverage at the seam.** "No behavior change" is the right constraint, but without tests the seam silently degrades. Existing OOP integration tests should pass *and* a new unit-level test for `OutOfProcessHost` lifecycle (start → ping → setup → shutdown → restart) should be added.

13. **Issue #4 (Benchmark harness) ordering claim is half-right.** The harness *infrastructure* (HttpListener, BenchmarkDotNet config, scenario definitions) can be built in parallel with #2 and #3. The harness *runs* require both prototypes complete. Recommend splitting #4 into 4a (infrastructure, parallel with #2/#3) and 4b (wire to executor implementations, blocked on #2 and #3). Keeps the critical path honest.

14. **Issue #6 (Adopt the winner) is fine but should explicitly preserve `SubprocessHostMode: "Single"` as a supported fallback for at least one release.** Removing the loser entirely on the same PR that adopts the winner is too aggressive given how new this is.

### Out of scope, but worth noting

15. **Cancellation** (your open question §6.4): agree it's a separate issue. One footnote: in Option B, killing a single host on a per-request timeout no longer kills "the server's only PowerShell" — it kills 1 of N. The current "kill subprocess on timeout" behavior is actually *better* in B than today, even without proper cancellation. Worth a sentence in the plan.

16. **Application Insights / observability** (spec 008): the benchmark harness should run with AI **disabled** to keep latency numbers clean. Add to §4 harness setup.

### Verdict

**Comment / approve direction with revisions.** The plan is sound enough to start issue #1 (extract `OutOfProcessHost`) immediately. Items 1, 2, 5, 6, 8, 9, and 13 should be reflected in the plan or the corresponding follow-up issue descriptions before #2/#3/#4 begin. Items 3, 4, 10, 11, 14, 15, 16 are nice-to-have refinements that can be folded in during prototype work.

Reviewed against current code: `OutOfProcessCommandExecutor.cs` (`_sendLock` semantics, `SendRequestAsync`, `_pending` correlation) and `oop-host.ps1` (handler ordering, ndjson framing, `Write-NdjsonResponse` flush behavior). The plan's description of current state matches reality.

usepowershell:
## Plan revised per Cubert + Farnsworth review (commit 0dbd2b7)

Both reviews are now folded into `specs/004-out-of-process-execution/runspace-pool-experiment-plan.md`. Summary of material changes:

### Factual corrections (Cubert)
- Dropped the `Phases 1-4 complete` framing — spec 004 has no phase manifest. Shipped capabilities now described directly.
- Added a path-drift footnote: Issue #65 references `specs/out-of-process-execution.md`; the canonical path is `specs/004-out-of-process-execution/`.
- Promoted the pre-existing single-runspace `\System.Management.Automation.ParseException: At line:20 char:37
+ **Operational note:** EMU policy on `usepowershell/PoshMcp` blocks bo …
+                                     ~~
The Unicode escape sequence is not valid. A valid sequence is `u{ followed by one to six hex digits and a closing '}'.

At line:20 char:163
+ … nt` from `stmuraws_microsoft`; reviews were posted via the `usepowers …
+                                                              ~~
The Unicode escape sequence is not valid. A valid sequence is `u{ followed by one to six hex digits and a closing '}'.
   at System.Management.Automation.Runspaces.PipelineBase.Invoke(IEnumerable input)
   at Microsoft.PowerShell.Executor.ExecuteCommandHelper(Pipeline tempPipeline, Exception& exceptionThrown, ExecutionOptions options)` non-clearing bug (verified at `oop-host.ps1:558-565`) to its own follow-up issue (#0) and called it out in §1. The pool work amplifies it but does not cause it.
- Pinned `HttpListener` binding to `127.0.0.1:0` to avoid Windows URL ACL pitfalls.
- Pinned baseline capture: same harness, same run, same machine — no pre-captured baseline.

### Architectural revisions (Farnsworth — must-fix items 1, 2, 5, 6, 8, 9, 13)
- **Item 1 (Option A stream hygiene):** corrected the `[Console]::Out` redirection claim. `Console.Out` is process-global and cannot be scoped per runspace via `InitialSessionState`. Replaced with a custom `PSHost` + `PSHostUserInterface` passed to `CreateRunspacePool` — that's the only realistic interception point. Defense-in-depth: `PSStyle PlainText` + `NO_COLOR=1` + the existing `IsNonJsonPowerShellStreamLine` backstop.
- **Item 2 (Option A setup race):** added an explicit quiesce protocol — drain pending → close pool → rebuild ISS → reopen pool → resume. Noted that an executor-side gate may also be needed since `SendRequestAsync` does not gate setup against in-flight invokes today.
- **Item 5 (Option B startup policy):** swapped blanket fail-fast for `first host fail-fast (smoke test), hosts 2..N retry with backoff`. Added `SubprocessMinHealthyForStartup` knob.
- **Item 6 (Option B crash-mid-lease race):** added `ConcurrentDictionary<int, HostState>` alongside the channel. Pool size is reconciled from the dictionary, not the channel. Replacement hosts are added to the channel only after `ping + setup` succeed.
- **Item 8 (Option A isolation gate):** split the crash-recovery scenario. Process-kill is the gate for B (same as baseline for A — not a useful A discriminator). Added a runspace-level corruption probe as the real isolation gate for A.
- **Item 9 (per-scenario thresholds):** replaced the single 4x bar with: ≥4x for I/O and network-shaped, ≥2x for heavy serialization, ≥1.5x for CPU-bound, parity (within 1.5x) for CPU-light.
- **Item 13 (#4 ordering):** split into #4a (harness infrastructure — parallel with #2/#3) and #4b (wiring — blocked on #2/#3).

### Refinements folded in (Farnsworth — items 3, 4, 10, 11, 14, 15, 16)
- **Item 3:** added per-pipeline stream guidance (`\.Streams.*`) beyond just `\System.Management.Automation.ParseException: At line:20 char:37
+ **Operational note:** EMU policy on `usepowershell/PoshMcp` blocks bo …
+                                     ~~
The Unicode escape sequence is not valid. A valid sequence is `u{ followed by one to six hex digits and a closing '}'.

At line:20 char:163
+ … nt` from `stmuraws_microsoft`; reviews were posted via the `usepowers …
+                                                              ~~
The Unicode escape sequence is not valid. A valid sequence is `u{ followed by one to six hex digits and a closing '}'.
   at System.Management.Automation.Runspaces.PipelineBase.Invoke(IEnumerable input)
   at Microsoft.PowerShell.Executor.ExecuteCommandHelper(Pipeline tempPipeline, Exception& exceptionThrown, ExecutionOptions options)`.
- **Item 4:** required pool queue-depth and lease-wait metrics in the Option A prototype.
- **Item 10:** parameterized the heavy-serialization scenario to `-First {10,100,1000}`.
- **Item 11:** captured in §6 Q3 — local `HttpListener` is sufficient; real-Azure validation is a follow-up acceptance test.
- **Item 14:** required `SubprocessHostMode: ""Single""` to remain a supported fallback for at least one full release after adoption (#6).
- **Item 15:** noted in §3 and §6 Q4 that Option B's per-request kill-on-timeout is structurally safer than today even before proper cancellation propagation.
- **Item 16:** required AI / spec-008 logging disabled during benchmark runs.

### Refinement deferred to a follow-up issue
- **Item 12:** Farnsworth asked for explicit unit-level test coverage on the #1 (extract `OutOfProcessHost`) seam. Added to the follow-up #1 description in the plan rather than building the test rig in this plan PR — the test belongs with the refactor, not with the planning doc.

A new `Revision History` section at the bottom of the plan summarizes the same changes.

PR #200, opened 2026-05-06
feat(oop): Option B prototype — process pool executor (#192)

usepowershell:
**🔧 Bender (Backend Developer)**

Closes #192

## Pool design

A pool of N independent `pwsh` subprocess hosts as an alternative to the single-subprocess executor. Each request leases one host, runs against it, and returns it when finished. Hosts that crash or time out are reconciled out of rotation and replaced by a background reconciler. Built on the `OutOfProcessHost` seam from #190.

## Lease + reconciliation

- `Channel<HostSlot>` for the available queue (lease = `ReadAsync`, return = `TryWrite`).
- `ConcurrentDictionary<int, HostSlot>` keyed by stable slot index for source-of-truth about which slots exist and their status.
- Both structures are required: the channel alone cannot detect a crashed mid-lease host; the dictionary lets the reconciler discover dead slots and start replacements without process-id reuse hazards.

## Fail-fast vs backoff

- **Slot 0 is fail-fast** — if the first host can''t reach a healthy `ping`, `StartAsync` throws.
- **Slots 1..N-1** start in parallel with bounded retry and exponential backoff.
- Pool starts as long as `SubprocessMinHealthyForStartup` hosts come up healthy (degraded if any failed); otherwise `StartAsync` throws and the pool is disposed.

## Discovery cache key

Per-pool cache keyed by a **SHA-256 fingerprint** over the `EnvironmentConfiguration` fields that influence module visibility: module paths, install/import modules, startup script path/content, `trustPSGallery`, and the effective discovery module list. Replacement hosts reapply the same setup, so the fingerprint and cached schemas remain valid across reconciliation. Setup with a different fingerprint invalidates the cache.

## Per-request kill on timeout

When `InvokeAsync` observes a `TimeoutException` (or a host dies mid-invoke), the pool marks that specific slot dead and the reconciler replaces it. The rest of the pool keeps serving — Option B''s structural advantage over single-host mode.

## Defaults

| Knob | Default |
|------|---------|
| `SubprocessHostMode` | `"Single"` |
| `SubprocessPoolSize` | `4` |
| `SubprocessMinHealthyForStartup` | `1` |
| StartupRetryCount | 3 |
| StartupBackoffInitial | 250 ms |
| StartupBackoffMax | 5 s |
| ReconcilerInterval | 1 s |
| Request timeout | 30 s |

## Test results

[...]

Issue #195, opened 2026-05-06
OOP: Run benchmarks and write findings


usepowershell:
## Summary

Execute the wired benchmark harness and produce findings. **Three runs per scenario per mode**, results table committed alongside a written analysis that picks a recommended mode.

## Acceptance criteria

- [ ] Three runs per scenario per mode; report mean and variance.
- [ ] Results table committed under `specs/004-out-of-process-execution/` (e.g., `benchmark-results.md`).
- [ ] Written analysis covers: cold-start, throughput, serialization vs payload size, crash-recovery (process and runspace), queue-depth / lease-wait behavior.
- [ ] Recommendation: which mode wins, why, and any caveats.
- [ ] Open questions from §6 of the plan addressed where the data permits (default `N`, A-vs-B keep-both decision, network-shaped fidelity, cancellation propagation scope).

Spec: specs/004-out-of-process-execution/spec.md
Plan: specs/004-out-of-process-execution/runspace-pool-experiment-plan.md (§5 #5)
Refs #65
Blocked by: #194

PR #201, opened 2026-05-06
feat(oop): Option A prototype — runspace pool host (#191)

### LINKED ISSUE #191: OOP: Option A prototype — runspace pool host (SubprocessHostMode: "Pool")
## Summary

Prototype Option A: a single `pwsh` subprocess that hosts a runspace pool, with `oop-host-pool.ps1` and a `SubprocessHostMode: "Pool"` config flag. Goal is to determine whether one process running N runspaces in parallel beats N processes for our workload.

## Acceptance criteria

- [ ] New `oop-host-pool.ps1` running an ISS-based pre-warmed runspace pool.
- [ ] **Synchronized writer** so concurrent runspaces never interleave ndjson frames on stdout.
- [ ] **Custom `PSHost` / `PSHostUserInterface`** for stream isolation (the previously planned `[Console]::Out` redirection is process-global and cannot scope per-runspace — do not use it).
- [ ] **Per-runspace ``** handling and **per-pipeline stream handling** via `PowerShell.Streams.*` (Verbose, Warning, Information, Debug) — not just ``.
- [ ] **Quiesce protocol for `setup`**: `setup` must drain in-flight invokes before mutating module state, then resume. `setup` must not race with concurrent `invoke`.
- [ ] **Queue-depth and lease-wait metrics** emitted from the prototype so Option A can be evaluated against the bench.
- [ ] `SubprocessHostMode: "Pool"` config flag toggles between `Single` (today) and `Pool`.
- [ ] Existing OOP integration tests pass against both modes.

Spec: specs/004-out-of-process-execution/spec.md
Plan: specs/004-out-of-process-execution/runspace-pool-experiment-plan.md (§5 #2)
Refs #65
Blocked by: #190

usepowershell:
**🛠️ Hermes (PowerShell Expert)**

Closes #191

## Design summary

Adds a runspace-pool variant of the OOP subprocess host so multiple invokes can execute concurrently inside a single `pwsh` process. Default behavior is unchanged: `SubprocessHostMode` defaults to `Single`.

- `oop-host-pool.ps1` speaks the same ndjson wire protocol as `oop-host.ps1` (`ping` / `setup` / `discover` / `invoke` / `shutdown`, request/response correlated by `id`). stdout is reserved for ndjson frames; diagnostics go to stderr.
- ISS-based pre-warmed runspace pool: `CreateRunspacePool(1, PoolSize, ISS, CustomHost)` with `ReuseThread` + MTA. `ImportPSModule` entries from setup are applied to the `InitialSessionState` so every leased runspace is warm.
- Custom `PSHost` (`NdjsonHost`) + `PSHostUserInterface` (`NdjsonHostUI`) route all `System.Management.Automation.Internal.Host.InternalHost.UI` writes (`Write-Host`, warnings, verbose, debug, progress) to stderr so cmdlet output cannot pollute the ndjson stdout channel. `[Console]::Out` is process-global and cannot be scoped per-runspace via `InitialSessionState` — a custom `PSHost` is the realistic intercept point.
- Synchronized writer: `PoolStdout.Lock` guards every complete ndjson frame write so concurrent worker responses never interleave on stdout.
- Per-pipeline stream isolation: errors / warnings / verbose / debug / information are read from `.Streams.*` (per-`PowerShell` instance), not from the runspace-wide `Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-191\PoshMcp.Server\PowerShell\OutOfProcess\oop-host-pool.ps1.new' because it does not exist. Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-192\PoshMcp.Server\PowerShell\OutOfProcess\oop-host-pool.ps1' because it does not exist. Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-192\PoshMcp.Server\PowerShell\OutOfProcess\oop-host-pool.ps1.new' because it does not exist. Cannot find path 'C:\Users\stmuraws\AppData\Local\Temp\poshmcp191-test-build.log' because it does not exist. Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-192\test-out.txt' because it does not exist.`. `Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-191\PoshMcp.Server\PowerShell\OutOfProcess\oop-host-pool.ps1.new' because it does not exist. Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-192\PoshMcp.Server\PowerShell\OutOfProcess\oop-host-pool.ps1' because it does not exist. Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-192\PoshMcp.Server\PowerShell\OutOfProcess\oop-host-pool.ps1.new' because it does not exist. Cannot find path 'C:\Users\stmuraws\AppData\Local\Temp\poshmcp191-test-build.log' because it does not exist. Cannot find path 'C:\Users\stmuraws\source\github\usepowershell\poshmcp-192\test-out.txt' because it does not exist.` is also explicitly cleared at the top of every user-script block so contamination from a prior invoke on the same leased runspace cannot leak.
- C# `PoolDispatcher` with N worker threads (= pool size). Each worker calls `ps.Invoke()` synchronously against a runspace leased from the pool, so true parallelism comes from the pool itself. A pure C# dispatcher avoids the PowerShell `ScriptBlock`-to-`Action` default-runspace deadlock that occurs when the host's main read loop pins the default runspace.

## Quiesce protocol

`setup` MUST NOT race with concurrent invokes. The pool host:

1. Resets `DrainEvent` so new invokes block before submission.
2. Calls `PoolDispatcher.WaitIdle` (60 s timeout) to let in-flight work complete. Timeout is logged as a warning; setup still proceeds.
3. Closes / disposes the existing runspace pool.
4. Applies module-path / install-module / import-module / startup-script mutations to the host process.
5. Rebuilds `InitialSessionState` from cached setup params and reopens the pool so all subsequent leases see the new state.
6. Sets `DrainEvent` to resume accepting invokes.

## Metrics emitted

Each invoke response carries a `metrics` object on the result frame:

```json
"metrics": {
  "queueDepthOnArrival": 0,
  "leaseWaitMs": 0,
  "activeOnComplete": 1,
  "poolSize": 4
}
```

- `queueDepthOnArrival` — pending + active observed when the work item was queued.
- `leaseWaitMs` — time the item spent waiting for a worker thread before `Invoke()`.
- `activeOnComplete` — active worker count at response time.
- `poolSize` — configured pool capacity.

Embedding metrics on the existing response frame (rather than a separate metrics frame) keeps the wire surface unchanged for clients and correlates metrics one-to-one with their request `id` without adding a second protocol channel.

## Pool size default + rationale

Default = `min(Environment.ProcessorCount, 8)`.

- Most module imports keep at least one core busy on dispatch + JSON serialization, so 1:1 with logical CPUs is a reasonable starting point.
- The cap at 8 prevents unbounded fan-out on high-core hosts during prototype evaluation.
- `PowerShellConfiguration.SubprocessRunspacePoolSize` overrides the default when set to a positive value.

## Configuration

- `PowerShellConfiguration.SubprocessHostMode`: `Single` (default) | `Pool`.
- `PowerShellConfiguration.SubprocessRunspacePoolSize`: `0` = host default.
- `McpToolSetupService` forwards both to `OutOfProcessCommandExecutor`; the executor selects `oop-host.ps1` vs `oop-host-pool.ps1` at start and forwards `runspacePoolSize` to `setup`.

## Test results

- Existing OOP integration tests run unchanged against `Single` mode.
- New `OutOfProcessPoolHostIntegrationTests` covers:
  - Pool start + discover.
  - Concurrent invokes (4 parallel `Start-Sleep`) complete well under the serial baseline, proving real parallelism.
  - Setup-during-invoke quiesce: a setup call after a batch of concurrent invokes drains, rebuilds ISS, reopens the pool, and accepts new invokes successfully.
  - Stream-isolation guard: `Write-Host` inside a tool body does not corrupt the ndjson channel.
- `dotnet test --filter "FullyQualifiedName~OutOfProcess"`: **112 passed, 0 failed**, 6 pre-existing skips.

### DIFF
--- a/PoshMcp.Server/PoshMcp.csproj
+++ b/PoshMcp.Server/PoshMcp.csproj
@@ -58,6 +58,7 @@
   <ItemGroup>
     <EmbeddedResource Include="default.appsettings.json" />
     <EmbeddedResource Include="PowerShell\OutOfProcess\oop-host.ps1" />
+    <EmbeddedResource Include="PowerShell\OutOfProcess\oop-host-pool.ps1" />
     <EmbeddedResource Include="..\infrastructure\azure\deploy.ps1" Link="Infrastructure\Azure\deploy.ps1" />
     <EmbeddedResource Include="..\infrastructure\azure\validate.ps1" Link="Infrastructure\Azure\validate.ps1" />
     <EmbeddedResource Include="..\infrastructure\azure\main.bicep" Link="Infrastructure\Azure\main.bicep" />
--- a/PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessCommandExecutor.cs
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessCommandExecutor.cs
@@ -30,6 +30,8 @@ public class OutOfProcessCommandExecutor : ICommandExecutor
     private readonly ILogger<OutOfProcessCommandExecutor> _logger;
     private readonly ILoggerFactory _loggerFactory;
     private readonly TimeSpan _requestTimeout;
+    private readonly SubprocessHostMode _hostMode;
+    private readonly int _poolSize;
 
     private OutOfProcessHost? _host;
     private bool _disposed;
@@ -42,13 +44,19 @@ public class OutOfProcessCommandExecutor : ICommandExecutor
     /// <param name="requestTimeout">
     /// Timeout for individual requests to the subprocess. Defaults to 30 seconds.
     /// </param>
+    /// <param name="hostMode">Selects the host script (Single or Pool). Defaults to Single.</param>
+    /// <param name="runspacePoolSize">Pool size when <paramref name="hostMode"/> is Pool. 0 lets the host pick a default.</param>
     public OutOfProcessCommandExecutor(
         ILogger<OutOfProcessCommandExecutor> logger,
-        TimeSpan? requestTimeout = null)
+        TimeSpan? requestTimeout = null,
+        SubprocessHostMode hostMode = SubprocessHostMode.Single,
+        int runspacePoolSize = 0)
     {
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         _loggerFactory = NullLoggerFactory.Instance;
         _requestTimeout = requestTimeout ?? TimeSpan.FromSeconds(30);
+        _hostMode = hostMode;
+        _poolSize = runspacePoolSize;
     }
 
     /// <summary>
@@ -58,13 +66,22 @@ public class OutOfProcessCommandExecutor : ICommandExecutor
     /// </summary>
     public OutOfProcessCommandExecutor(
         ILoggerFactory loggerFactory,
-        TimeSpan? requestTimeout = null)
+        TimeSpan? requestTimeout = null,
+        SubprocessHostMode hostMode = SubprocessHostMode.Single,
+        int runspacePoolSize = 0)
     {
         ArgumentNullException.ThrowIfNull(loggerFactory);
         _loggerFactory = loggerFactory;
         _logger = loggerFactory.CreateLogger<OutOfProcessCommandExecutor>();
         _requestTimeout = requestTimeout ?? TimeSpan.FromSeconds(30);
+        _hostMode = hostMode;
+        _poolSize = runspacePoolSize;
     }
+
+    /// <summary>
+    /// The host mode this executor was constructed with.
+    /// </summary>
+    public SubprocessHostMode HostMode => _hostMode;
 
     /// <inheritdoc />
     public async Task StartAsync(CancellationToken cancellationToken = default)
@@ -188,6 +205,8 @@ public class OutOfProcessCommandExecutor : ICommandExecutor
             skipPublisherCheck = config.SkipPublisherCheck,
             allowClobber = config.AllowClobber,
             installTimeoutSeconds = config.InstallTimeoutSeconds,
+            // Pool host reads this to size its runspace pool. Ignored by single-runspace host.
+            runspacePoolSize = _poolSize,
         };
 
         _logger.LogInformation("Sending environment setup to OOP subprocess.");
@@ -306,17 +325,21 @@ public class OutOfProcessCommandExecutor : ICommandExecutor
     }
 
     /// <summary>
-    /// Resolves the path to the oop-host.ps1 script.
+    /// Resolves the path to the host script for the configured <see cref="HostMode"/>.
     /// Priority: environment variable override → embedded resource extraction → build output fallback.
     /// </summary>
     internal async Task<string> ResolveHostScriptPathAsync()
     {
+        var scriptName = _hostMode == SubprocessHostMode.Pool
+            ? "oop-host-pool.ps1"
+            : "oop-host.ps1";
+
         var overridePath = Environment.GetEnvironmentVariable("POSHMCP_OOP_HOST_PATH");
         if (!string.IsNullOrWhiteSpace(overridePath))
         {
             if (File.Exists(overridePath))
             {
-                _logger.LogInformation("Using override oop-host.ps1 from POSHMCP_OOP_HOST_PATH: {Path}", overridePath);
+                _logger.LogInformation("Using override OOP host script from POSHMCP_OOP_HOST_PATH: {Path}", overridePath);
                 return overridePath;
             }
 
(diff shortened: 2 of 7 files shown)


usepowershell:

**APPROVED**

Reviewed against `specs/004-out-of-process-execution/runspace-pool-experiment-plan.md` §5 #2 and the PR's Hermes-criteria checklist. The implementation correctly addresses every architectural concern raised against the plan, including the `[Console]::Out`-is-process-global correction and the quiesce protocol that the plan revision flagged as required.

---

### Criteria verification

1. **ISS-based pre-warmed runspace pool.** `New-PoolInitialSessionState` builds an `InitialSessionState` from `CreateDefault2()` with `ThreadOptions = ReuseThread`, calls `ImportPSModule` for each `importModules` entry, and `Open-Pool` constructs `[runspacefactory]::CreateRunspacePool(1, $script:PoolSize, $iss, $script:PoolHost)` with `ApartmentState = MTA`. Modules are bound at ISS time so every leased runspace is warm. ✅

2. **Synchronized stdout writer.** `PoolStdout.Lock` (a single `object`) gates every `Console.Out.WriteLine + Flush` call. Both the PowerShell-side `Write-NdjsonResponse` and the C#-side `PoolDispatcher.ProcessOne` route through `PoolStdout.Write(string)`, so concurrent worker responses cannot interleave. ✅

3. **Custom `PSHost` / `PSHostUserInterface`, not `[Console]::Out`.** `NdjsonHost` + `NdjsonHostUI` + `NdjsonHostRawUI` are passed to `CreateRunspacePool`. Every `$Host.UI` write path (`Write`, `WriteLine`, `WriteDebugLine`, `WriteVerboseLine`, `WriteWarningLine`, `WriteErrorLine`, `WriteProgress`) routes to `Console.Error` with a typed prefix; prompts throw `NotSupportedException`. The header comment correctly explains why `[Console]::Out` cannot be scoped per-runspace. ✅

4. **Per-pipeline streams + per-runspace `$Error` clear.** `PoolDispatcher.ProcessOne` reads `w.Ps.Streams.Error` and `w.Ps.Streams.Warning` directly off the `PowerShell` instance and uses `w.Ps.HadErrors` for the `hadErrors` flag — never runspace-wide `$Error`. The `$userScript` body in `Invoke-InvokeHandler` calls `$Error.Clear()` before invoking the user command, eliminating contamination from a prior invoke that landed on the same leased runspace. ✅

5. **Quiesce protocol for `setup`.** `Begin-Drain` does `DrainEvent.Reset()` then `PoolDispatcher.WaitIdle(60000)`; on timeout it logs a warning and proceeds (documented tradeoff). After drain, `Close-Pool` tears down the existing pool, mutations are applied to the host process, `Ensure-Pool` rebuilds the ISS and reopens the pool, then `End-Drain` sets `DrainEvent`. `Invoke-InvokeHandler` calls `$script:DrainEvent.Wait()` before submitting to the dispatcher, so new invokes block until quiesce ends. ✅

6. **Queue-depth and lease-wait metrics.** Every invoke response carries a `metrics` object on the result frame: `queueDepthOnArrival`, `leaseWaitMs`, `activeOnComplete`, `poolSize`. Choice (response frame vs separate frame) is documented in the PR description and in the script header. Wire surface is unchanged for `Single` clients. ✅

7. **`SubprocessHostMode` config flag, default `Single`.** `PowerShellConfiguration.SubprocessHostMode` is an `enum SubprocessHostMode { Single, Pool }` with default `Single`. `SubprocessRunspacePoolSize` (default 0 → host picks `min(ProcessorCount, 8)`) is forwarded through `OutOfProcessCommandExecutor` and embedded in the `setup` request payload. `ResolveHostScriptPathAsync` parameterizes the script name (`oop-host.ps1` vs `oop-host-pool.ps1`). ✅

8. **Existing OOP integration tests run against both modes.** Single-mode path in `OutOfProcessCommandExecutor` is unchanged on the hot path; the new ctor parameters default to `Single`/`0`. PR description reports `dotnet test --filter "FullyQualifiedName~OutOfProcess"` = 112 passed, 0 failed. Confirmed in worktree. ✅

9. **New integration tests cover concurrent invokes (≥4) and setup-during-invoke quiesce.** `OutOfProcessPoolHostIntegrationTests.PoolHost_ConcurrentInvokesRunInParallel` runs 4 parallel `Start-Sleep 500ms` and asserts elapsed < 0.75 × serial baseline — proves real parallelism without flakiness. `PoolHost_SetupQuiesceCompletesAfterInvokes` runs a batch, calls `SetupAsync` (drains, rebuilds ISS, reopens), then runs another batch. `PoolHost_StreamIsolation_WriteHostDoesNotPolluteStdout` confirms the custom-host intercept holds. ✅

---

### Observations (non-blocking)

- **Drain timeout is hardcoded at 60 s.** `Begin-Drain` calls `WaitIdle(60000)` and on timeout proceeds anyway with a warning. This is documented but not configurable; a future operator with `SubprocessTimeoutSeconds = 120` would expect drain to honor that. Worth a follow-up to thread the executor's request timeout (or a dedicated `SubprocessDrainTimeoutSeconds`) into the host script's `setup` payload.
- **`Invoke-DiscoverHandler` does not `DrainEvent.Wait()`.** Discover runs synchronously on the read thread, so it cannot overlap with the read loop. But if `setup` were re-entered before the read thread returned to the loop (it isn't, but the invariant relies on stdin being single-reader), discover could observe a closed pool. Not a defect today; worth a comment to prevent future regressions.
- **Inline C# block embeds ~250 lines of `Add-Type` code.** Acknowledged in the agent decision note. Maintainability tradeoff is accepted for the prototype phase. If this mode wins per #195/#196, consider lifting the C# helper to a small native assembly co-located with the host script and loaded via `[Reflection.Assembly]::LoadFrom`.
- **Pool size cap of 8 is a prototype guard.** Reasonable for unbounded fan-out protection on high-core hosts. Should be removed (or raised significantly) if Option A wins, since I/O-bound Az workloads scale beyond `ProcessorCount`.
- **`Resolve-SwitchParameters` runs on the read thread before submission.** This calls `Get-Command` in the host process (not in a runspace), which depends on whatever modules the host process has imported. For switch parameters defined in modules imported only into the pool runspaces (via ISS), this lookup may fail or return ambiguous results. Worth confirming that `ImportPSModule` on ISS makes the cmdlet visible to the host process's `Get-Command`. If not, switch resolution should happen inside the user script.

---

### Cross-PR coordination — `SubprocessHostMode` collision (#200 vs #201)

Both this PR and #200 (Bender / Option B) introduce a type named `SubprocessHostMode`. They are **not source-compatible**:

- This PR (#201): `enum SubprocessHostMode { Single, Pool }`. Property is `SubprocessHostMode SubprocessHostMode`.
- PR #200: `static class SubprocessHostMode` with string constants `Single`/`ProcessPool`/`Pool`. Property is `string? SubprocessHostMode`.

Whichever PR lands second will need a non-trivial rebase. This is a coordination issue, not a defect. My recommendation is to standardize on this PR's `enum` (stronger typing, idiomatic C#) and extend it with `ProcessPool` when #200 rebases. The rebase touches `PowerShellConfiguration.SubprocessHostMode` (string? → enum), `McpToolSetupService.IsProcessPool(...)` (string check → enum equality on `SubprocessHostMode.ProcessPool`), and Bender's unit tests covering options validation. Estimated rebase scope: ~30 lines.

---

### Verdict

**APPROVED.** All nine evaluation criteria are met, the architectural concerns from the plan revision (custom `PSHost` for stream isolation, quiesce protocol for `setup`, per-pipeline streams, per-runspace `$Error`) are addressed correctly, and the parallelism and quiesce tests provide meaningful guards. Land this when ready; coordinate the enum convergence with #200 on whichever lands second.

Issue #196, opened 2026-05-06
OOP: Adopt the winner — make recommended mode default


usepowershell:
## Summary

Adopt the winning mode from #5 as the new default; demote the loser to opt-in. Update spec 004 and `DOCKER.md`.

## Acceptance criteria

- [ ] Recommended mode becomes the default `SubprocessHostMode`.
- [ ] **`SubprocessHostMode: "Single"` MUST remain a supported fallback for at least one full release.** (Removing it on the same PR that adopts the winner is too aggressive given how new this work is.)
- [ ] Spec 004 updated to reflect the chosen architecture.
- [ ] If Option A wins: explicitly note that the "subprocess isolation protects the server" contract from US2 weakens at the runspace level.
- [ ] `DOCKER.md` updated for any new env vars / config defaults.
- [ ] CHANGELOG entry under the appropriate release.

Spec: specs/004-out-of-process-execution/spec.md
Plan: specs/004-out-of-process-execution/runspace-pool-experiment-plan.md (§5 #6)
Refs #65
Blocked by: #195

usepowershell:
Closed by PR #208 (default flip + doctor + docs) and PR #209 (run-4 reaffirmation: Pool/Single 4.87x, gate sustained). Both merged to main.

PR #202, opened 2026-05-06
feat(benchmarks): wire harness to executors (#194)

### LINKED ISSUE #194: OOP: Wire benchmark harness to executors
## Summary

Connect the benchmark harness from #4a to all three executors: `OutOfProcessCommandExecutor` (today's `Single`), the runspace-pool host (Option A), and `OutOfProcessSubprocessPool` (Option B). Blocks on both prototypes existing.

## Acceptance criteria

- [ ] Harness exercises `Single`, `Pool`, and `ProcessPool` modes from one run.
- [ ] All scenarios from #4a executable end-to-end against each executor.
- [ ] AI / spec-008 logging disabled during runs.
- [ ] Output Markdown table includes mode, scenario, payload size, mean, p95, p99, crash-recovery time.
- [ ] Documented invocation (`dotnet run -c Release ...`) so #5 can be executed reproducibly.

Spec: specs/004-out-of-process-execution/spec.md
Plan: specs/004-out-of-process-execution/runspace-pool-experiment-plan.md (§5 #4b)
Refs #65
Blocked by: #191, #192, #193

usepowershell:
**🧪 Fry (Tester)**

Closes #194

## What got wired

- **ExecutorFactory** — builds ICommandExecutor per HostMode using the same construction path the production server uses (OutOfProcessCommandExecutor + OutOfProcessSubprocessPool), without the hosting/configuration/logging stack. Skips `SetupAsync` because built-in cmdlets are callable without environment customization — the harness measures executor cost, not module install.
- **HostMode [Params] axis** on every scenario (Single, Pool, ProcessPool) — one `dotnet run` produces one results table covering all three configurations.
- **All five scenarios** wired end-to-end: `ColdStartBenchmark`, `WarmInvokeThroughputBenchmark`, `PayloadSizeSerializationBenchmark`, `ProcessCrashRecoveryBenchmark`, `RunspaceCorruptionRecoveryBenchmark`.
- **Crash injection** lives in `ExecutorFactory` via reflection into private fields — keeps the bench-only hook out of production types. `InternalsVisibleTo` for `PoshMcp.Benchmarks` added to `PoshMcp.csproj` for the few internal types referenced through the public `ICommandExecutor` surface.
- **P99StatisticColumn** custom BDN column — the AC requires P99 and BenchmarkDotNet 0.14 only ships `StatisticColumn.P95`.
- **AI / spec-008 logging disabled** inside the bench process — `Program.cs` sets `ApplicationInsights__Enabled=false` and clears `APPLICATIONINSIGHTS_CONNECTION_STRING` before BDN starts.

## Output columns (per AC)

| Column         | Source |
|----------------|--------|
| Mode           | `HostMode` `[Params]` |
| Scenario       | one row per `[Benchmark]` method |
| Payload size   | `PayloadBytes` `[Params]` (256 KB / 1 MB rows on the size scenario) |
| Mean           | BDN built-in |
| P95            | `StatisticColumn.P95` |
| P99            | `P99StatisticColumn` (custom) |
| Crash-recovery | `Mean` column on `ProcessCrashRecoveryBenchmark` rows IS the per-mode recovery time |

## Invocation (full reproduction lives in PoshMcp.Benchmarks/README.md)

```powershell
# All scenarios, all modes
dotnet run --project PoshMcp.Benchmarks --configuration Release -- --filter *

# One scenario across all modes
dotnet run --project PoshMcp.Benchmarks --configuration Release -- --filter '*WarmInvoke*'

# Filter by mode
dotnet run --project PoshMcp.Benchmarks --configuration Release -- --filter '*Mode=ProcessPool*'

# Smoke test (near-zero iterations)
dotnet run --project PoshMcp.Benchmarks --configuration Release -- --filter '*ColdStart*' --job dry
```

Markdown output lands in `PoshMcp.Benchmarks/BenchmarkDotNet.Artifacts/` — the `*-report-github.md` file is what #195 will consume.

## Smoke-test results

- **Build (Release):** 0 errors, no new warnings (2 pre-existing `NU1510` from `System.Security.Cryptography.Xml` in `PoshMcp.csproj`).
- **`--list flat`:** all 5 scenarios discovered.
- **`ColdStartBenchmark --job dry`:** PASS for `Single` (5.5 s), `Pool` (5.9 s), `ProcessPool` (6.0 s) — confirms the harness drives all three executors end-to-end.
- **`WarmInvokeThroughputBenchmark --job dry`:** PASS for `Pool` (265–318 ms / 10 calls). `Single` and `ProcessPool` throw `OOP error: An item with the same key has already been added. Key: Content` from `OutOfProcessHost.SendRequestAsync` when 10 concurrent invokes share a single host. **This is a real production bug in `OutOfProcessHost` that the harness exposes — not a bench-wiring defect.** Filed as a follow-up; does not block #194 (the harness wiring is correct, and the failure mode is exactly what a benchmark is supposed to surface).
- **Full regression suite (Debug):** 629 passed, 0 failed, 7 skipped.

## Expected full-run duration

A complete run across all five scenarios × three modes × the payload-size axis is on the order of **30–60 minutes** on a modern dev box; allow longer when including the 1 MB payload row and accounting for repeated cold-start launches in `ProcessCrashRecoveryBenchmark`.

### DIFF
--- a/PoshMcp.Benchmarks/BenchmarkConfig.cs
+++ b/PoshMcp.Benchmarks/BenchmarkConfig.cs
@@ -1,6 +1,8 @@
+using BenchmarkDotNet.Columns;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Exporters;
 using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Reports;
 
 namespace PoshMcp.Benchmarks;
 
@@ -26,5 +28,18 @@ public sealed class BenchmarkConfig : ManualConfig
         // BenchmarkDotNet so we don't have to enumerate them here. Scenarios
         // that need per-scenario tuning use attributes on the [Benchmark] methods.
         Add(DefaultConfig.Instance);
+
+        // Surface percentile columns called out in the AC for #194:
+        // "Markdown table includes columns: mode, scenario, payload size,
+        //  mean, p95, p99, crash-recovery time".
+        // BDN ships a static StatisticColumn.P95 but not P99 — see
+        // P99StatisticColumn for the custom column we add to fill the gap.
+        // For crash-recovery scenarios the Mean column IS the recovery time
+        // (each iteration kills + recovers); documented in
+        // PoshMcp.Benchmarks/README.md.
+        AddColumn(StatisticColumn.P95);
+        AddColumn(new P99StatisticColumn());
+
+        SummaryStyle = SummaryStyle.Default.WithRatioStyle(RatioStyle.Trend);
     }
 }
--- a/PoshMcp.Benchmarks/ExecutorFactory.cs
+++ b/PoshMcp.Benchmarks/ExecutorFactory.cs
@@ -1,0 +1,229 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging.Abstractions;
+using PoshMcp.Server.PowerShell.OutOfProcess;
+
+namespace PoshMcp.Benchmarks;
+
+/// <summary>
+/// Builds an <see cref="ICommandExecutor"/> for each <see cref="HostMode"/>
+/// using the same wiring path the production server uses
+/// (<c>McpToolSetupService.StartOutOfProcessExecutorIfNeededAsync</c>),
+/// without pulling in the hosting / configuration / logging stack.
+///
+/// <para>
+/// Skip the executor's <c>SetupAsync</c> step on purpose: benchmarks measure
+/// the executor itself, not environment customization (module install /
+/// import / startup script). Built-in cmdlets like <c>Get-Date</c>,
+/// <c>Get-Random</c>, and <c>Write-Output</c> are callable via
+/// <see cref="ICommandExecutor.InvokeAsync"/> without prior discovery.
+/// </para>
+/// </summary>
+internal static class ExecutorFactory
+{
+    /// <summary>
+    /// Default size for the <see cref="HostMode.ProcessPool"/> pool.
+    /// Matches <c>PowerShellConfiguration.SubprocessPoolSize</c>'s default.
+    /// </summary>
+    public const int DefaultProcessPoolSize = 4;
+
+    /// <summary>
+    /// Default size for the <see cref="HostMode.Pool"/> runspace pool.
+    /// 0 lets the host pick min(ProcessorCount, 8).
+    /// </summary>
+    public const int DefaultRunspacePoolSize = 0;
+
+    /// <summary>
+    /// Constructs and starts an executor for the given <paramref name="mode"/>.
+    /// Caller is responsible for <see cref="ICommandExecutor.DisposeAsync"/>.
+    /// </summary>
+    public static async Task<BenchExecutor> CreateAsync(
+        HostMode mode,
+        TimeSpan? requestTimeout = null,
+        CancellationToken cancellationToken = default)
+    {
+        var timeout = requestTimeout ?? TimeSpan.FromSeconds(30);
+
+        switch (mode)
+        {
+            case HostMode.Single:
+            {
+                var exec = new OutOfProcessCommandExecutor(
+                    NullLoggerFactory.Instance,
+                    requestTimeout: timeout,
+                    hostMode: SubprocessHostMode.Single,
+                    runspacePoolSize: 0);
+                await exec.StartAsync(cancellationToken).ConfigureAwait(false);
+                return new SingleOrPoolBenchExecutor(exec);
+            }
+
+            case HostMode.Pool:
+            {
+                var exec = new OutOfProcessCommandExecutor(
+                    NullLoggerFactory.Instance,
+                    requestTimeout: timeout,
+                    hostMode: SubprocessHostMode.Pool,
+                    runspacePoolSize: DefaultRunspacePoolSize);
+                await exec.StartAsync(cancellationToken).ConfigureAwait(false);
+                return new SingleOrPoolBenchExecutor(exec);
+            }
+
+            case HostMode.ProcessPool:
+            {
+                // Resolve pwsh + the single-runspace host script (ProcessPool
+                // hosts run oop-host.ps1, not the pool variant — each subprocess
+                // owns one runspace).
+                var resolver = new OutOfProcessCommandExecutor(
+                    NullLoggerFactory.Instance,
+                    hostMode: SubprocessHostMode.Single);
+                var hostScript = await resolver.ResolveHostScriptPathAsync().ConfigureAwait(false);
+                var pwshPath = OutOfProcessCommandExecutor.ResolvePwshPath();
+                await resolver.DisposeAsync().ConfigureAwait(false);
+
+                var pool = new OutOfProcessSubprocessPool(
+                    pwshPath,
+                    hostScript,
+                    new OutOfProcessSubprocessPoolOptions
+                    {
+                        PoolSize = DefaultProcessPoolSize,
+                        MinHealthyForStartup = 1,
+                    },
+                    NullLoggerFactory.Instance,
+                    requestTimeout: timeout);
+
+                await pool.StartAsync(cancellationToken).ConfigureAwait(false);
+                return new ProcessPoolBenchExecutor(pool);
+            }
+
+            default:
+                throw new ArgumentOutOfRangeException(nameof(mode), mode, "Unknown HostMode.");
+        }
+    }
+}
+
+/// <summary>
+/// Wraps an <see cref="ICommandExecutor"/> with crash-injection helpers the
+/// benchmark scenarios need but that aren't part of the production interface.
+/// </summary>
+internal abstract class BenchExecutor : IAsyncDisposable
+{
+    public abstract ICommandExecutor Executor { get; }
+
+    /// <summary>
+    /// Forcibly kills one underlying pwsh subprocess to simulate a crash.
+    /// Returns the killed PID, or null if no live host could be found.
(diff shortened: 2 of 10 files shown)


usepowershell:

REQUEST CHANGES — one hard blocker (CI is red on a formatting check, not an architectural defect). Architecture is sound; once CI is green this flips to approve.

## Verdict summary

| Area | Status |
|------|--------|
| AC: harness exercises Single / Pool / ProcessPool from one run | ✅ — `[Params(HostMode.Single, HostMode.Pool, HostMode.ProcessPool)]` on every scenario class. |
| AC: all five scenarios end-to-end (no stubs) | ✅ — Cold-start, Warm-invoke, Payload-size, ProcessCrashRecovery, RunspaceCorruptionRecovery all wired. No `NotImplementedException`. |
| AC: AI / spec-008 logging disabled at process start | ✅ — `Program.cs` clears `ApplicationInsights__Enabled` and `APPLICATIONINSIGHTS_CONNECTION_STRING` before `BenchmarkSwitcher.Run`. Bench harness also doesn't construct `IHost` / DI, so spec-008 init never executes — env-var clear is belt-and-suspenders. |
| AC: Markdown columns mode / scenario / payload / mean / p95 / p99 / crash-recovery | ✅ — `BenchmarkConfig` adds `StatisticColumn.P95` + custom `P99StatisticColumn`; `Mode` and `PayloadBytes` come from `[Params]`; crash-recovery time = `Mean` on `ProcessCrashRecoveryBenchmark`. |
| AC: README documents reproducible invocation | ✅ — `PoshMcp.Benchmarks/README.md` covers run, filter-by-scenario, filter-by-mode, smoke job, and output location. |
| CI status | ❌ — `build / Verify formatting` failed (run 25465584796). |

## 🔴 Blocker — CI `dotnet format --verify-no-changes` fails

The `build` job fails with ~50 `error WHITESPACE: Fix whitespace formatting. Insert '\s\s\s\s'` errors in `PoshMcp.Benchmarks/ExecutorFactory.cs` lines 54–100. The switch-case bodies need an extra indentation level per the repo's `.editorconfig`. This is a hard merge blocker on the CI gate independent of any code review.

**Fix:** `dotnet format PoshMcp.sln` from repo root, commit the whitespace-only diff, push.

## 🏗️ Architecture review

### `ExecutorFactory` — construction parity ✅

Wires each `HostMode` through the same constructor surface the production server uses (`OutOfProcessCommandExecutor` for Single/Pool, `OutOfProcessSubprocessPool` for ProcessPool). Skipping `SetupAsync` is the right call — the harness measures executor cost, not module install — and the README documents this explicitly. Built-in cmdlets (`Get-Date`, `Write-Output`, `Start-Sleep`, `Invoke-WebRequest`) are all callable without discovery, so the choice is sound.

One small construction quirk worth noting (non-blocking): for `ProcessPool`, the factory instantiates a throw-away `OutOfProcessCommandExecutor` purely to call `ResolveHostScriptPathAsync()`, then disposes it. Cheap, but exposing `ResolveHostScriptPathAsync` as a static or via a small resolver helper would let the bench skip that allocation. Not a merge blocker.

### Reflection-based crash injection ✅ acceptable, with fragility note

`KillOneHost()` reaches into `OutOfProcessCommandExecutor._host`, `OutOfProcessHost._process`, and `OutOfProcessSubprocessPool._slots` via private-field reflection. This is correctly:
- Bench-only (`PoshMcp.Benchmarks` assembly, never compiled into Server),
- Clearly labeled in XML doc comments as a benchmark contract loose against future refactors,
- Failure-safe (`return null` on missing fields → caller gets a null PID).

**Fragility:** if any of those private field names are renamed, `KillOneHost` silently returns `null`, the iteration's "next invoke" succeeds against the still-live host, and `ProcessCrashRecoveryBenchmark` quietly degrades to "warm-invoke latency" without anyone noticing. Suggest a one-line guard at startup that asserts the reflection lookups resolve, and aborts the run if any field is gone. Defer to a follow-up.

### `InternalsVisibleTo` for `PoshMcp.Benchmarks` ✅ minimal

Diff adds exactly one IVT entry to `PoshMcp.csproj` for `PoshMcp.Benchmarks`. The bench needs internal types (`OutOfProcessCommandExecutor`, `OutOfProcessHost`, `OutOfProcessSubprocessPool`, `OutOfProcessSubprocessPoolOptions`, `SubprocessHostMode`) to construct executors with the same shape the production wiring uses. Scoped to one assembly. Not a blanket open-up. ✅

### `P99StatisticColumn` ✅ — built-in really doesn't exist

Confirmed: BDN 0.14 ships `StatisticColumn.P95` but not `P99`; `WithPercentiles` is per-summary and `Statistics.Percentiles` exposes individual properties only via `ResultStatistics`. The custom column walks `ResultStatistics.OriginalValues`, sorts once, applies linear-interpolated rank (R-7), and renders in ms. Implementation is straightforward and matches the magnitude of other time columns. No built-in alternative was missed.

### AI-logging disable ✅ sufficient

Two layers of defense:
1. `Program.cs` sets `ApplicationInsights__Enabled=false` and clears the connection string env var before BDN starts.
2. The bench harness builds no `IHost`, no DI container, no `IConfiguration` — the spec-008 init code path (`AddApplicationInsightsTelemetry` / OTel setup) is never reached because nobody constructs `WebApplication.CreateBuilder` or `Host.CreateApplicationBuilder`.

Could spec-008 still emit through another path? The only realistic concern would be the `ApplicationInsights*` SDK auto-initializing from a global env var, but the env-var clear handles that. No `Activity` / OTel exporter is registered without DI. ✅

### Crash-recovery on `Mean` column — acceptable AC interpretation

The AC reads "Markdown table includes columns: ... crash-recovery time." Implementation puts the recovery time in the standard `Mean` column on `ProcessCrashRecoveryBenchmark` rows (one kill+invoke per iteration via `[InvocationCount(1)]`). For `ProcessPool`, this is the genuine recovery metric — kill 1 of N hosts, lease loop skips the dead slot, next invoke lands on a healthy host, Mean = real recovery time. For `Single`/`Pool`, the iteration disposes the executor and reconstructs (the only host is dead) — Mean reports cold-start cost, which is honestly labeled in both code comments and README as "the correct answer to 'time until next successful request' for those modes."

This isn't a literal "Crash-Recovery (ms)" header but the README table explicitly maps the AC column to the BDN column, and the per-mode semantics are documented. Acceptable as written.

## 🟡 Non-blocking observations

1. **`RunspaceCorruptionRecoveryBenchmark` measures head-of-line blocking, not corruption.** The class name says "corruption recovery" but the body fires a 1-second `Start-Sleep` and times a fast `Get-Date` probe. Documented design rationale (process-kill is the wrong gate for Option A) is sound — but the AC text literally says "runspace-corruption recovery." Suggest renaming the class to `HeadOfLineBlockingBenchmark` or `RunspaceContentionBenchmark` so the deliverable name matches what's measured. Also, "corruption" implies state damage; this measures contention. Defer to a small follow-up rename PR.

2. **`PayloadSizeSerializationBenchmark` — 1 MB row × ProcessPool will be slow** and may fight the bumped 60 s timeout under load. Already noted in the PR body's 30–60 minute estimate. Worth tagging the 1 MB params row with a longer `[Timeout]` attribute or splitting it into a separate scenario class so the small-payload runs aren't held up. Defer.

3. **`WarmInvokeThroughputBenchmark` reuses one HTTP server across iterations.** Fine for the scenario, but note the listener uses a dynamic port — concurrent BDN runs against this assembly will collide on `HttpListenerTestServer` initialization. Mostly a "don't run the benchmark twice in parallel" caveat, not a defect.

## ⚖️ The `OutOfProcessHost.SendRequestAsync` "Key: Content" bug

PR body acknowledges that for `Single` / `ProcessPool` the warm-invoke smoke fails with `An item with the same key has already been added. Key: Content` from `OutOfProcessHost.SendRequestAsync` under 10 concurrent invokes through one host. ProcessPool hits this because each pool host has one runspace and concurrent invokes routed within a single host serialize through the same `_pending` dict — so the production bug is a real concurrency race in the host itself, not a harness defect.

**My call: this does NOT block #194.** Reasoning:

- The deliverable for #194 is *the harness is wired to all three executors*, not *the harness produces clean numbers on day one*. AC #1 ("exercises Single, Pool, AND ProcessPool from one run") is satisfied — the cold-start smoke passes for all three modes (5.5 s / 5.9 s / 6.0 s). The wiring is provably correct.
- A benchmark that surfaces a real production race on its first concurrent run is doing the job a benchmark exists to do. Suppressing the failure to land #194 cleanly would defeat the purpose.
- The fix lives in production code (`OutOfProcessHost`), not in the harness. Coupling it to #194 inflates this PR's scope and delays #195 / #196.
- The Pool executor (Hermes / Option A, PR #201) does NOT hit the bug because its dispatcher is concurrent-aware. That's actually useful comparative data — it argues *for* Option A on this dimension.

File the bug as a separate `spec:004` / `bug` issue (with the harness commit / scenario as the repro recipe) and let #194 land once CI is green. The harness's exposure of the bug should be referenced from the new issue as the regression gate.

## What needs to happen before merge

1. **Required:** Run `dotnet format PoshMcp.sln`, commit the whitespace fixes, push. CI must go green.
2. **Required:** File the `OutOfProcessHost` concurrency bug as a separate tracked issue (spec:004) and reference it from this PR's body before merge so the failure mode is on the record.
3. **Optional / deferrable:** Reflection-resolution startup assertion in `ExecutorFactory.KillOneHost()`. Rename `RunspaceCorruptionRecoveryBenchmark` → `HeadOfLineBlockingBenchmark`. Both as small follow-ups, not blockers.

Once item 1 lands (CI green), this is APPROVED — the architecture is right, the AC is met, the surfaced production bug is the harness doing its job.

— Farnsworth

usepowershell:
**🧪 Fry (Tester)** Format fix pushed at 87fb244; CI should be green now.

PR #204, opened 2026-05-06
fix(oop): SendRequestAsync 'Key: Content' under parallel invokes (#203)

### LINKED ISSUE #203: OOP: Concurrency bug — OutOfProcessHost.SendRequestAsync 'Key: Content' under parallel invokes
**🧪 Fry (Tester)** — surfaced this while wiring the benchmark harness in #194 / PR #202.

## Summary

`OutOfProcessHost.SendRequestAsync` throws `An item with the same key has already been added. Key: Content` when multiple concurrent invokes are issued against a single host. Reproduces deterministically in `WarmInvokeThroughputBenchmark` (10 parallel invokes / 1 host) for both `SubprocessHostMode = Single` and `SubprocessHostMode = ProcessPool` (where individual slots can each receive concurrent invokes if the lease is reused improperly, or where the failure surfaces under the bench's per-mode parallelism).

`SubprocessHostMode = Pool` does NOT fail — runs at 265–318 ms / 10 invokes — because the runspace-pool host script in `oop-host-pool.ps1` already serializes ndjson frames behind `PoolStdout.Lock` AND distributes invokes across pre-warmed runspaces.

## Where it surfaces

```
OutOfProcessHost.SendRequestAsync  →  System.ArgumentException:
  An item with the same key has already been added. Key: Content
```

The exception originates inside the `Dictionary<string, ...>` add path used to build the request frame (likely the request body / headers dictionary in `SendRequestAsync`), implying the same `Content` key is being added twice when two threads enter the build path concurrently. This is a textbook unsynchronized-dictionary-mutation bug.

## Why this matters

- `Single` mode is the default. Any deployment that fans concurrent MCP `tools/call` requests into a single subprocess host hits this.
- `ProcessPool` mode also hits it whenever a single slot serves more than one in-flight invoke (lease-reuse path or pool exhaustion).
- The `Pool` path (runspace pool inside one process) does not exhibit it because of the lock around the ndjson writer.

## Repro

From [PR #202](https://github.com/usepowershell/PoshMcp/pull/202):

```powershell
dotnet run --project PoshMcp.Benchmarks --configuration Release -- `
  --filter '*WarmInvoke*' --job dry
```

`Single` and `ProcessPool` rows fail with the above message; `Pool` succeeds.

## Acceptance criteria

- [ ] `OutOfProcessHost.SendRequestAsync` is safe under concurrent callers (either by serializing access around the request-frame build + write under `_sendLock`, or by ensuring per-call dictionaries are local to the call).
- [ ] Add a unit test that fires N concurrent `SendRequestAsync` calls against one host and asserts no `ArgumentException` and that all responses correlate correctly via the `_pending` map.
- [ ] `WarmInvokeThroughputBenchmark` runs clean for `Single` and `ProcessPool` modes after the fix.
- [ ] Existing OOP integration tests continue to pass.

## Refs

- Surfaced via #194 / PR #202
- Spec: `specs/004-out-of-process-execution/spec.md`
- Related: #190 (the `OutOfProcessHost` extraction), #200 (process pool consumer)

usepowershell:
**🔧 Bender (Backend Developer)**

Closes #203.

## Actual root cause

The issue body hypothesized a C# dictionary race in `OutOfProcessHost` (the request/response correlation map being mutated under concurrent `SendRequestAsync` calls). **That hypothesis was wrong.**

The error — `OOP error: An item with the same key has already been added. Key: Content` — originates inside the PowerShell host, not the C# host. `ConvertTo-Json -Depth 4` enumerates the reflection members of the object being serialized. `Invoke-WebRequest -UseBasicParsing` returns a `BasicHtmlWebResponseObject`, whose `Content` property CLR-shadows the base `WebResponseObject.Content`. `ConvertTo-Json` sees two members named `Content` and throws `System.ArgumentException` from the underlying `Dictionary<string, object>`.

It is not concurrency-related. The `WarmInvokeThroughputBenchmark` harness fires N parallel `Invoke-WebRequest` invokes, and that pattern made the failure deterministic, but a single invoke would also trip it.

## Fix

In `PoshMcp.Server/PowerShell/OutOfProcess/oop-host.ps1` and the embedded user-script in `PoshMcp.Server/PowerShell/OutOfProcess/oop-host-pool.ps1`, the `ConvertTo-Json` call is wrapped in a layered fallback:

1. **Primary:** `ConvertTo-Json -Depth 4 -Compress` directly. Cheap and correct for almost all objects.
2. **Fallback A — flatten via PSObject:** If primary throws, pipe the value through `Select-Object *`. This materializes a flat `PSObject` that collapses shadowed CLR members into a single property. Retry `ConvertTo-Json`.
3. **Fallback B — string representation:** If even Fallback A throws (unusual — e.g. cyclic graphs), emit the object's `ToString()` as a JSON string so the client receives valid JSON rather than an exception bleed.

The fallback is invisible to the C# client on the happy path. The C# correlation map in `OutOfProcessHost` is untouched by this PR — it was never the source of the bug.

## Regression test

New: `PoshMcp.Tests/Unit/OutOfProcess/OutOfProcessHostConcurrencyTests.cs`

- `InvokeAsync_ConcurrentInvokeWebRequest_DoesNotThrowDuplicateKeyError` — fires 10 parallel `Invoke-WebRequest` calls against an ephemeral-port loopback `HttpListener`. Pre-fix this throws `System.InvalidOperationException` with `OOP error: An item with the same key has already been added. Key: Content`. Post-fix all 10 calls return non-empty JSON.
- `SendRequestAsync_ConcurrentCallers_AllResponsesCorrelate` — fires 20 parallel `ping` requests and asserts each task receives `status=ok`. Sanity net for the correlation map (validates the original hypothesis was wrong, and protects against a future real race).

Both tests skip cleanly on systems without `pwsh` or `HttpListener`.

## Smoke verify

`dotnet build PoshMcp.sln -c Release` clean (0 errors).
`dotnet run --project PoshMcp.Benchmarks -c Release -- --filter '*WarmInvoke*' --job dry` — all three modes complete without the `Key: Content` error:

| Mode        | Concurrency | Mean     |
|-------------|-------------|----------|
| Single      | 10          | 781.8 ms |
| Pool        | 10          | 306.0 ms |
| ProcessPool | 10          | 357.9 ms |

Pre-fix, `Single` and `ProcessPool` failed with the duplicate-key error inside the bench's invoke path; `Pool` was already green because `oop-host-pool.ps1` happened to be working through a slightly different code path that didn't hit the same `Invoke-WebRequest` pattern in the harness's repro. Post-fix all three are clean.

## Files changed

- `PoshMcp.Server/PowerShell/OutOfProcess/oop-host.ps1` — JSON serialization fallback
- `PoshMcp.Server/PowerShell/OutOfProcess/oop-host-pool.ps1` — same fallback in the embedded user-script
- `PoshMcp.Tests/Unit/OutOfProcess/OutOfProcessHostConcurrencyTests.cs` — regression coverage (new file)

### DIFF
--- a/PoshMcp.Server/PowerShell/OutOfProcess/oop-host-pool.ps1
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/oop-host-pool.ps1
@@ -872,11 +872,26 @@ function Invoke-InvokeHandler {
     # (cleared so contamination from a prior invoke on the same runspace
     # cannot leak). The pipeline pre-serializes the result to JSON so the
     # C# dispatcher can embed it without a second round-trip into PowerShell.
+    # The ConvertTo-Json call is wrapped in try/catch with a Select-Object *
+    # fallback to tolerate objects whose CLR type shadows a base-class
+    # member of the same name (e.g. BasicHtmlWebResponseObject's 'Content'
+    # shadows WebResponseObject.Content). See issue #203.
     $userScript = {
         param($Name, $Splat)
         $Error.Clear()
         $r = & $Name @Splat
-        return ($r | ConvertTo-Json -Depth 4 -Compress -WarningAction SilentlyContinue)
+        if ($null -eq $r) { return 'null' }
+        try {
+            return ($r | ConvertTo-Json -Depth 4 -Compress -WarningAction SilentlyContinue)
+        }
+        catch [System.ArgumentException] {
+            try {
+                return ($r | Select-Object * | ConvertTo-Json -Depth 4 -Compress -WarningAction SilentlyContinue)
+            }
+            catch {
+                return (($r | Out-String).Trim() | ConvertTo-Json -Compress)
+            }
+        }
     }
     [void]$ps.AddScript($userScript)
     [void]$ps.AddArgument($commandName)
--- a/PoshMcp.Server/PowerShell/OutOfProcess/oop-host.ps1
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/oop-host.ps1
@@ -54,6 +54,45 @@ function Write-NdjsonResponse {
     [Console]::Out.Flush()
 }
 
+function ConvertTo-SafeJson {
+    <#
+    .SYNOPSIS
+        Serialize an object to compact JSON, tolerating duplicate-property
+        errors thrown by ConvertTo-Json on objects whose CLR type shadows a
+        base-class member of the same name (e.g. BasicHtmlWebResponseObject's
+        'Content' shadows WebResponseObject.Content). Falls back to a
+        Select-Object * projection that materializes a flat PSObject (which
+        de-duplicates shadowed members), then to a string representation.
+    .NOTES
+        See issue #203. Without this wrapper, `Invoke-WebRequest |
+        ConvertTo-Json -Depth 4` throws
+        `An item with the same key has already been added. Key: Content`,
+        which surfaces to the C# client as an OOP error.
+    #>
+    param(
+        [Parameter(Mandatory)][AllowNull()]$InputObject,
+        [int]$Depth = 4
+    )
+
+    if ($null -eq $InputObject) { return 'null' }
+
+    try {
+        return $InputObject | ConvertTo-Json -Depth $Depth -Compress -WarningAction SilentlyContinue 3>$null
+    }
+    catch [System.ArgumentException] {
+        # Duplicate property name (shadowed member). Project to a flat
+        # PSObject so PowerShell's member resolver collapses duplicates.
+        try {
+            $projected = $InputObject | Select-Object *
+            return $projected | ConvertTo-Json -Depth $Depth -Compress -WarningAction SilentlyContinue 3>$null
+        }
+        catch {
+            # Last resort: stringify and JSON-encode the string.
+            return (($InputObject | Out-String).Trim() | ConvertTo-Json -Compress)
+        }
+    }
+}
+
 function Invoke-PingHandler {
     <#
     .SYNOPSIS
@@ -570,7 +609,9 @@ function Invoke-InvokeHandler {
             $hadErrors = $true
         }
 
-        $jsonOutput = $result | ConvertTo-Json -Depth 4 -Compress -WarningAction SilentlyContinue 3>$null
+        # Use ConvertTo-SafeJson to tolerate shadowed-property objects
+        # (e.g. Invoke-WebRequest results). See issue #203.
+        $jsonOutput = ConvertTo-SafeJson -InputObject $result -Depth 4
         if ($null -eq $jsonOutput) {
             $jsonOutput = 'null'
         }
--- a/PoshMcp.Tests/Unit/OutOfProcess/OutOfProcessHostConcurrencyTests.cs
+++ b/PoshMcp.Tests/Unit/OutOfProcess/OutOfProcessHostConcurrencyTests.cs
@@ -1,0 +1,257 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Sockets;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging.Abstractions;
+using PoshMcp.Server.PowerShell.OutOfProcess;
+using Xunit;
+
+namespace PoshMcp.Tests.Unit.OutOfProcess;
+
+/// <summary>
+/// Concurrency tests for <see cref="OutOfProcessHost.SendRequestAsync{T}"/>.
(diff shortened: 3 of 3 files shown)


usepowershell:

**Verdict: APPROVED**

## Root cause assessment

Diagnosis holds. `BasicHtmlWebResponseObject` (returned by `Invoke-WebRequest -UseBasicParsing`) declares its own `Content` (a string for the response body) which CLR-shadows the inherited `WebResponseObject.Content` (the raw `byte[]` stream). `ConvertTo-Json` walks reflection members and inserts each into a `Dictionary<string, object>`; the shadowed pair collides on the second `Add`, throwing `System.ArgumentException: An item with the same key has already been added. Key: Content`. This is a well-known PowerShell serialization edge case, not a concurrency bug — the harness's parallel `Invoke-WebRequest` pattern just made the failure deterministic. The PR body's correction of the original hypothesis is accurate and clearly stated.

The C# correlation map (`_pending` in `OutOfProcessHost`) was correctly identified as a red herring and is untouched. The second regression test (`SendRequestAsync_ConcurrentCallers_AllResponsesCorrelate`) is a sensible sanity net for that original hypothesis even though it isn't the failing path.

## Fix targeting & scope

- `oop-host.ps1`: Wrapper extracted as `ConvertTo-SafeJson`, applied at the single failing site (`Invoke-InvokeHandler` line ~611) where the user-command result is serialized. Other `ConvertTo-Json` call sites in the file (request-frame envelopes, error frames) are not wrapped — correct, since those serialize controlled C#-shaped payloads, not arbitrary user output.
- `oop-host-pool.ps1`: Same fallback inlined into the user-script scriptblock. Inlining (rather than calling a host-process helper) is the right call here — the scriptblock executes inside a pooled runspace and shouldn't depend on host-process function availability. Asymmetry between the two files is intentional and acceptable.
- Fallback only triggers on `catch [System.ArgumentException]`; happy path is unchanged. No risk of silent corruption for callers whose objects serialize cleanly today.

## Fallback semantics

- **Tier 1 (`Select-Object *`):** Materializes a flat `PSObject` whose member view collapses shadowed CLR members (derived wins). For `BasicHtmlWebResponseObject` this means the string `Content` — the body — wins over the `byte[]` shadow. That is the property MCP callers actually want, so this is the correct projection.
- **Tier 2 (`($r | Out-String).Trim() | ConvertTo-Json`):** Last-resort string repr. This IS semantic loss (clients lose structure) but only fires after `Select-Object *` itself throws `ArgumentException` — realistically only on cyclic/exotic graphs. Returning a valid JSON string instead of bleeding an exception to the C# client is the right tradeoff at the bottom of the chain.

The escape hatch is bounded: shadowed-member objects (the common case) hit Tier 1 and return structured JSON; pathological objects hit Tier 2 and return a string. Either way, the C# client never sees the duplicate-key error.

## Regression test adequacy

`InvokeAsync_ConcurrentInvokeWebRequest_DoesNotThrowDuplicateKeyError` exercises the exact failure mode: real `Invoke-WebRequest -UseBasicParsing` against a loopback `HttpListener`, which produces a real `BasicHtmlWebResponseObject`. Concurrency=10 mirrors the harness repro. Pre-fix this throws `OOP error: An item with the same key has already been added. Key: Content`; post-fix it asserts each response carries a non-empty `output`. Skip guards on `pwsh` and `HttpListener.IsSupported` are correct. The script-resolution helper (embedded resource → fingerprinted temp file → `BaseDirectory` fallback) is well-factored.

One small note (non-blocking): the test only covers `oop-host.ps1` (single-host path). The pool-host inline fallback in `oop-host-pool.ps1` is structurally equivalent but not directly exercised by these tests. The smoke result on `WarmInvokeThroughputBenchmark` Pool mode (306 ms / 10 calls clean) is reasonable end-to-end coverage; a dedicated pool-mode unit test could be a follow-up if Bender wants belt-and-suspenders, but I don't consider it blocking.

## Concurrency framing

PR body is explicit: *"It is not concurrency-related."* Good. The branch name (`squad/203-host-concurrency-fix`) and the issue title still carry the original "concurrency" framing — that's fine because they predate the diagnosis, and the PR body sets the record straight for future readers walking the history.

## Cross-cutting / sequencing note (#195)

This PR unblocks `WarmInvokeThroughputBenchmark` for `Single` and `ProcessPool` modes. Hermes is currently running PR #195 (benchmarks + findings doc) on a parallel branch and has captured runs 1+2 against pre-#203 `main`, where Single/ProcessPool throughput numbers are unreliable (the benchmark was hitting the duplicate-key error inside the invoke loop). Once #204 lands, Hermes should rebase #195 onto the post-#203 `main` and rerun the affected benchmarks to capture clean Single/ProcessPool throughput numbers before publishing findings. **Not blocking #204** — flagging for the post-merge sequencing.

## Recommendation

Approve and merge. Convert this comment to a formal GitHub Approve review if branch protection requires it (EMU policy on this account blocks `gh pr review`, hence the comment-channel verdict).

PR #205, opened 2026-05-07
bench(oop): canonical results + findings (#195)

usepowershell:
Closes #195

## What's in this PR

- `specs/004-out-of-process-execution/benchmark-results.md` — canonical run-3 results (post-#204), all three `HostMode` values across ColdStart, PayloadSize, and WarmInvoke.
- `specs/004-out-of-process-execution/benchmark-findings.md` — interpretation per scenario + adoption recommendation for #196.
- `bench-runs/run-3.log` and `bench-runs/run-3-artifacts/` — raw BenchmarkDotNet output (csv / html / github md).

## Headline numbers (run-3, `--job short`)

| Scenario | Winner | Margin |
|---|---|---|
| WarmInvoke @ conc=10 | **Pool** | 4.86× vs Single (mean), 4.79× (P99); ProcessPool 3.30× / 3.41× |
| ColdStart | **Single** | ~400–500 ms ahead of Pool / ProcessPool |
| PayloadSize ≤16 KB | **Single** | per-invoke overhead dominates |
| PayloadSize @256 KB | **ProcessPool** | tightest mean & tail at this size |
| PayloadSize @1 MB | **Pool** | lowest allocations (~13.8 MB vs ~16.3 / ~17.4) |

ProcessPool also posts the tightest StdDev / P99 of the three on WarmInvoke (1.11 ms / 201.4 ms).

## Recommendation TL;DR

Default `HostMode` should flip to **Pool** (Option A). ProcessPool stays as opt-in for tail-sensitive / isolation-sensitive workloads. Single stays as documented choice for short-lived CLI invocations. Full justification, caveats, and open questions in [`benchmark-findings.md` § 4](specs/004-out-of-process-execution/benchmark-findings.md#4-recommendation-for-196).

## Sequencing

This PR closes #195 and **unblocks #196** (adopt winner). #196 owns the actual default flip, the doc sweep, the opt-in story, and the cancellation-as-prerequisite gating decision called out in the findings caveats.

## Caveats (from § 5)

- `--job short` is fast, not exhaustive — no SLO claims off these numbers.
- Single Windows 11 / Arm64 host; x64 not measured.
- Pool's trust boundary is weaker than ProcessPool's; custom `PSHost` work from PR #187 review remains a prerequisite for adversarial workloads.
- Cancellation propagation is not exercised here; #196 should treat it as a gate, not a follow-up.

Issue #189, opened 2026-05-06
OOP: Bug-fix — clear $Error before invoke in single-runspace host


usepowershell:
## Summary

In `oop-host.ps1`, `Invoke-InvokeHandler` (~lines 540-600) sets `hadErrors` from `.Count` (lines 561-565) without clearing `` before invoking the requested command. Because `` is process-global in the single-runspace host, errors from a prior invoke leak into the next invoke's `hadErrors` flag, producing false positives.

This is independent of the runspace-pool / process-pool experiment and can land any time.

## Acceptance criteria

- [ ] Add a one-line `.Clear()` before the command is invoked in `Invoke-InvokeHandler`.
- [ ] Add a regression integration test that performs two `invoke` calls in sequence: the first deliberately fails, the second succeeds. Assert the second response has `hadErrors == false`.
- [ ] Existing OOP integration tests continue to pass.

Spec: specs/004-out-of-process-execution/spec.md
Plan: specs/004-out-of-process-execution/runspace-pool-experiment-plan.md (§5 #0)
Refs #65

PR #207, opened 2026-05-07
feat(oop): cancellation propagation (#188)

### LINKED ISSUE #188: Cancellation propagation: cancelling the .NET wait does not stop in-flight pwsh work
## Problem

`CancellationToken` plumbed through the PowerShell execution path only cancels the **.NET-side wait** for a result. It does **not** stop the in-flight work executing inside `pwsh` — neither in the in-process runspace nor in the out-of-process (OOP) host subprocess.

When a caller cancels (timeout, client disconnect, shutdown), the .NET task completes promptly with `OperationCanceledException`/`TimeoutException`, but the underlying `PowerShell` pipeline keeps running to completion (or until the runspace/process is torn down). This wastes resources, holds locks/runspaces, and can produce ghost work that completes after the caller has moved on.

### Out-of-process path (`PoshMcp.Server/PowerShell/OutOfProcess/`)

In `OutOfProcessCommandExecutor.SendRequestAsync` (around L361–L432), the `cancellationToken` only governs:

- `_sendLock.WaitAsync(cancellationToken)`
- `_stdin.WriteLineAsync(..., cancellationToken)` / `FlushAsync(cancellationToken)`
- A linked `timeoutCts` that fails the local `TaskCompletionSource` with `TimeoutException`

The token is **never forwarded to the subprocess**. There is no `cancel` JSON-RPC method, and the host script (`oop-host.ps1`) processes requests in a **single-threaded dispatcher loop** (L630–L650). While `Invoke-InvokeHandler` (L498) is blocked inside `& $cmdInfo @boundParams`, the host cannot even read a follow-up `cancel` message from stdin — the loop is busy.

Net effect: cancellation in .NET releases the awaiter but the pwsh subprocess continues running the cmdlet to completion. The next request waits behind it.

### In-process path (`PoshMcp.Server/PowerShell/`)

`IsolatedPowerShellRunspace.ExecuteThreadSafeAsync` does not even accept a `CancellationToken`. The only place `_powerShell.Stop()` is called is in `Dispose()` (L146) when the runspace is being torn down. A caller cancellation does not translate into `PowerShell.Stop()` / `StopAsync()` on the live pipeline.

## Impact

- **Resource waste:** long-running cmdlets keep consuming CPU/memory/network after the caller is gone.
- **Head-of-line blocking (OOP):** because the host loop is single-threaded, a stuck `invoke` blocks every subsequent request on that subprocess until it finishes or the process is killed.
- **Misleading semantics:** the API surface advertises a `CancellationToken`, but cancellation is effectively cooperative-with-nothing. Callers reasonably assume cancelling the token stops the work.
- **Shutdown latency:** graceful shutdown can hang waiting for in-flight pwsh work that nothing is signaling to stop.
- **Test flakiness / leaks:** tests that cancel and move on can leave background pwsh work mutating shared state.

## Expected behavior

When a caller cancels the `CancellationToken` passed into a PowerShell invocation:

1. The in-flight pipeline is signaled to stop **as promptly as PowerShell allows** (cooperative `Stop()`/`StopAsync()` for in-process; a real cancel signal to the OOP host for OOP).
2. The .NET task completes with `OperationCanceledException` only **after** the underlying pipeline has been signaled (and ideally observed to have stopped or been forcibly terminated within a bounded grace period).
3. The runspace / OOP subprocess is left in a usable state for the next request, OR is recycled deterministically if it cannot be cleanly stopped.
4. Cancellation is observable in logs/metrics (which path was used: cooperative stop, forced stop, process kill).

## Possible approaches

These are not mutually exclusive — a layered solution is likely.

### In-process runspace
- Capture the `PSPowerShell` instance for the active invocation and register `cancellationToken.Register(() => ps.Stop())` (or `StopAsync()` on PS 7+) so cancellation triggers a cooperative pipeline stop.
- Update `IPowerShellRunspace` / `ExecuteThreadSafeAsync` signatures to accept a `CancellationToken`.
- Define a bounded grace window: if the pipeline does not stop within N seconds, mark the runspace poisoned and recycle it (existing `PowerShellCleanupService` is a natural home).

### Out-of-process host
- Add a `cancel` JSON-RPC method keyed by request `id`. The .NET side sends it from the cancellation callback.
- Make the host dispatcher capable of receiving `cancel` while an `invoke` is running. Options:
  - Run `invoke` on a background runspace/job inside the host while the dispatcher loop continues to read stdin.
  - Use a side channel (e.g., a named pipe or an OS signal) dedicated to control messages.
- Inside the host, translate `cancel` into `$psInstance.Stop()` / `Stop-Job` / etc. against the running pipeline.
- Define escalation: cooperative cancel → forced stop → kill subprocess and recycle, with bounded timeouts at each step. Process kill must be the last resort, not the default, because it loses module/runspace state.

### Cross-cutting
- Settle on whether cancellation should **always** be best-effort cooperative (with subprocess kill as backstop) or whether some operations are explicitly non-cancellable.
- Wire structured logging + metrics: cancellation requested, cancellation method used (cooperative/forced/kill), elapsed time to stop.
- Document the contract clearly in `IPowerShellRunspace` and `ICommandExecutor` XML docs.

## Acceptance criteria

- [ ] Cancelling a `CancellationToken` passed to in-process invocation calls `PowerShell.Stop()` / `StopAsync()` on the live pipeline within the cancellation callback.
- [ ] Cancelling a `CancellationToken` passed to OOP `InvokeAsync` causes the host subprocess to receive a cancel signal and attempt to stop the in-flight pipeline.
- [ ] OOP host can receive and act on a cancel message while an `invoke` is in progress (no head-of-line blocking on the dispatcher loop).
- [ ] If cooperative cancel does not complete within a configurable grace period, the runspace/subprocess is recycled deterministically and the next request gets a healthy executor.
- [ ] `IPowerShellRunspace.ExecuteThreadSafeAsync` (and any analogous in-process surface) accepts a `CancellationToken`.
- [ ] Logs/metrics record cancellation lifecycle: requested → signaled → stopped (cooperative | forced | killed) with timings.
- [ ] Integration tests cover: cancel during a long-running cmdlet (in-process and OOP), cancel of a stuck pipeline forcing recycle, and that a subsequent request on the same executor succeeds promptly after a cancellation.
- [ ] Public XML docs on the cancellation parameter accurately describe the new contract.

## References

- `PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessCommandExecutor.cs` — `SendRequestAsync`, `InvokeAsync`
- `PoshMcp.Server/PowerShell/OutOfProcess/oop-host.ps1` — single-threaded dispatcher loop, `Invoke-InvokeHandler`
- `PoshMcp.Server/PowerShell/OutOfProcess/ICommandExecutor.cs` — current `CancellationToken` contract
- `PoshMcp.Server/PowerShell/PowerShellRunspaceImplementations.cs` — in-process runspace, `_powerShell.Stop()` only on dispose
- `PoshMcp.Server/PowerShell/IPowerShellRunspace.cs` — async surface lacks `CancellationToken`
- `PoshMcp.Server/PowerShell/PowerShellCleanupService.cs` — likely home for recycle/escalation
- `specs/004-out-of-process-execution/` — original OOP execution spec; this issue extends its lifecycle/cancellation story

usepowershell:
Closes #188

Propagates `CancellationToken` cancellation through `OutOfProcessHost.SendRequestAsync` to the pwsh subprocess via a new `cancel` wire frame, so the in-flight pipeline is signaled to stop with `PowerShell.BeginStop()` instead of running to completion.

Full design: [`specs/004-out-of-process-execution/cancellation-design.md`](https://github.com/usepowershell/PoshMcp/blob/squad/188-cancellation-propagation/specs/004-out-of-process-execution/cancellation-design.md) (already on this branch, commit `9a5cf24`).

## Per-mode behavior

| Mode | What cancellation does |
|------|------------------------|
| `Single` (`oop-host.ps1`) | Refactored handler runs invokes on a background dispatcher thread against a shared runspace; `cancel` calls `BeginStop` on the matching `[powershell]` instance. Awaiter throws `OperationCanceledException` promptly; host stays healthy for follow-ups. |
| `Pool` (`oop-host-pool.ps1`) | `PoolDispatcher` tracks active items by request id; `Cancel(requestId)` calls `BeginStop` on the worker's `[powershell]`. No head-of-line blocking — concurrent invokes on other runspaces continue. |
| `ProcessPool` (`OutOfProcessSubprocessPool`) | Inherits soft-cancel for free via `OutOfProcessHost`. Existing per-request kill-on-timeout backstop preserved verbatim — wedged pipelines (unmanaged code) still get killed. |

## Test results

- 3 new cancellation tests (one per mode) in `PoshMcp.Tests/Unit/OutOfProcess/OutOfProcessCancellationTests.cs`:
  - `Single_LongRunningInvoke_TokenCancelStopsPipeline`
  - `Pool_LongRunningInvoke_TokenCancelStopsPipelineNoHeadOfLine`
  - `ProcessPool_TokenCancel_StopsPipelineAndKeepsSlotsHealthy`
- Full OOP test suite (`--filter "FullyQualifiedName~OutOfProcess"`): **148 passed, 0 failed, 6 skipped** (skips are Az module integration tests gated on environment).

## Wire protocol additions

- New `cancel` request frame: `{"id":"cancel-<guid>","method":"cancel","params":{"requestId":"<original-id>"}}`
- New optional `cancelled` boolean on invoke response frames.
- Cancel frame ids use `cancel-` prefix so unknown-id ack logs are downgraded to Debug.

## Sequencing

Unblocks #196 (Farnsworth — adopt-default flip to `Pool`).

### DIFF
--- a/PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessHost.cs
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/OutOfProcessHost.cs
@@ -238,14 +238,33 @@ public sealed class OutOfProcessHost : IAsyncDisposable
                 _sendLock.Release();
             }
 
-            using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            using var timeoutCts = new CancellationTokenSource();
             var requestTimeout = requestTimeoutOverride ?? _requestTimeout;
             timeoutCts.CancelAfter(requestTimeout);
 
-            var registration = timeoutCts.Token.Register(() =>
+            // Per-request timeout: notify the host so its pipeline stops, then trip the TCS.
+            // The cancel frame is best-effort and runs on its own short-lived CTS.
+            var timeoutRegistration = timeoutCts.Token.Register(() =>
             {
+                if (!"cancel".Equals(method, StringComparison.Ordinal))
+                {
+                    _ = TrySendCancelFrameAsync(id);
+                }
                 tcs.TrySetException(new TimeoutException(
                     $"Request {id} (method={method}) timed out after {requestTimeout.TotalSeconds}s."));
+            });
+
+            // Caller cancellation: forward to the host as a cancel frame so the in-flight
+            // pipeline is signaled to stop. The awaiter completes promptly with OCE; the
+            // host's eventual cancelled-response is dropped silently because _pending no
+            // longer has the id (see ReadLoopAsync).
+            var cancelRegistration = cancellationToken.Register(() =>
+            {
+                if (!"cancel".Equals(method, StringComparison.Ordinal))
+                {
+                    _ = TrySendCancelFrameAsync(id);
+                }
+                tcs.TrySetCanceled(cancellationToken);
             });
 
             try
@@ -262,12 +281,63 @@ public sealed class OutOfProcessHost : IAsyncDisposable
             }
             finally
             {
-                await registration.DisposeAsync().ConfigureAwait(false);
+                await timeoutRegistration.DisposeAsync().ConfigureAwait(false);
+                await cancelRegistration.DisposeAsync().ConfigureAwait(false);
             }
         }
         finally
         {
             _pending.TryRemove(id, out _);
+        }
+    }
+
+    /// <summary>
+    /// Best-effort: send a fire-and-forget <c>cancel</c> frame to the host so it can
+    /// signal the in-flight pipeline for <paramref name="requestId"/> to stop.
+    /// The frame id uses a <c>cancel-</c> prefix and is intentionally not registered
+    /// in <see cref="_pending"/>; the host's ack is logged at Debug only.
+    /// </summary>
+    private async Task TrySendCancelFrameAsync(string requestId)
+    {
+        if (_disposed || _stdin is null) return;
+
+        var cancelFrameId = "cancel-" + Guid.NewGuid().ToString("N");
+        var frame = new
+        {
+            id = cancelFrameId,
+            method = "cancel",
+            @params = new { requestId }
+        };
+
+        string json;
+        try
+        {
+            json = JsonSerializer.Serialize(frame);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogDebug(ex, "Failed to serialize cancel frame for request {Id}.", requestId);
+            return;
+        }
+
+        try
+        {
+            using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(2));
+            await _sendLock.WaitAsync(cts.Token).ConfigureAwait(false);
+            try
+            {
+                await _stdin.WriteLineAsync(json.AsMemory(), cts.Token).ConfigureAwait(false);
+                await _stdin.FlushAsync(cts.Token).ConfigureAwait(false);
+                _logger.LogDebug("Sent cancel frame for request {Id}.", requestId);
+            }
+            finally
+            {
+                try { _sendLock.Release(); } catch (ObjectDisposedException) { /* shutting down */ }
+            }
+        }
+        catch (Exception ex)
+        {
+            _logger.LogDebug(ex, "Failed to send cancel frame for request {Id}.", requestId);
         }
     }
 
@@ -379,7 +449,20 @@ public sealed class OutOfProcessHost : IAsyncDisposable
 
                     if (!_pending.TryGetValue(id, out var tcs))
                     {
-                        _logger.LogWarning("OOP response for unknown request id '{Id}': {Line}", id, line);
+                        // Cancel-frame acks (id prefix "cancel-") are intentionally
+                        // not registered; a normal invoke response that lost its race
+                        // with caller cancellation is also expected. Both are noise.
+                        if (id.StartsWith("cancel-", StringComparison.Ordinal)
+                            || (root.TryGetProperty("result", out var resForUnknown)
+                                && resForUnknown.TryGetProperty("cancelled", out var cancelledFlag)
+                                && cancelledFlag.ValueKind == JsonValueKind.True))
+                        {
+                            _logger.LogDebug("OOP response for already-removed request id '{Id}'.", id);
+                        }
+                        else
+                        {
+                            _logger.LogWarning("OOP response for unknown request id '{Id}': {Line}", id, line);
+                        }
                         continue;
                     }
 
--- a/PoshMcp.Server/PowerShell/OutOfProcess/oop-host-pool.ps1
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/oop-host-pool.ps1
@@ -208,6 +208,7 @@ namespace PoshMcp.PoolHost
         public int QueueDepthOnArrival;
         public Stopwatch QueueSw;
         public string CommandName;
+        public bool Cancelled; // set by Cancel() before BeginStop fires
     }
 
     public sealed class PoolDispatcher : IDisposable
@@ -217,6 +218,7 @@ namespace PoshMcp.PoolHost
         private readonly Thread[] _workers;
         private readonly ManualResetEventSlim _idle = new ManualResetEventSlim(true);
         private readonly object _idleLock = new object();
+        private readonly ConcurrentDictionary<string, PoolWorkItem> _active = new ConcurrentDictionary<string, PoolWorkItem>();
         private int _activeCount;
         private int _pendingCount; // queued + active, used for IsIdle
 
@@ -260,8 +262,33 @@ namespace PoshMcp.PoolHost
                 _idle.Reset();
             }
 
+            _active[id] = item;
             _queue.Add(item);
             return depth;
+        }
+
+        /// <summary>
+        /// Signal a cancellation request to the in-flight invoke matching
+        /// <paramref name="requestId"/>. Returns true if the request was found
+        /// and BeginStop was called; false if no matching active item exists
+        /// (already completed, queued-only, or unknown id).
+        /// </summary>
+        public bool Cancel(string requestId)
+        {
+            if (string.IsNullOrEmpty(requestId)) return false;
+            PoolWorkItem item;
+            if (!_active.TryGetValue(requestId, out item)) return false;
+            try
+            {
+                item.Cancelled = true;
+                item.Ps.BeginStop(null, null);
+                return true;
+            }
+            catch (Exception ex)
+            {
+                Console.Error.WriteLine("[oop-host-pool:dispatcher] BeginStop failed for " + requestId + ": " + ex.Message);
+                return false;
+            }
         }
 
         /// <summary>Wait until the dispatcher is idle (queue empty + no active work).</summary>
@@ -292,6 +319,8 @@ namespace PoshMcp.PoolHost
                     finally
                     {
                         try { w.Ps.Dispose(); } catch { }
+                        PoolWorkItem _removed;
+                        _active.TryRemove(w.Id, out _removed);
                         Interlocked.Decrement(ref _activeCount);
                         lock (_idleLock)
                         {
@@ -314,29 +343,51 @@ namespace PoshMcp.PoolHost
         {
             Collection<PSObject> output = null;
             string invokeError = null;
+            bool wasStopped = false;
 
             try
             {
                 output = w.Ps.Invoke();
+            }
+            catch (System.Management.Automation.PipelineStoppedException)
+            {
+                wasStopped = true;
             }
             catch (Exception ex)
             {
                 invokeError = ex.Message;
             }
 
+            // BeginStop transitions the pipeline to Stopped without throwing
+            // PipelineStoppedException for some host configurations; double-check.
+            if (!wasStopped)
+            {
+                try
+                {
+                    var state = w.Ps.InvocationStateInfo;
+                    if (state != null && state.State == PSInvocationState.Stopped)
+                    {
+                        wasStopped = true;
+                    }
+                }
+                catch { /* best-effort */ }
+            }
+
+            bool cancelled = wasStopped || w.Cancelled;
+
             string[] errs = w.Ps.Streams.Error.Select(e => e.ToString()).ToArray();
             string[] warns = w.Ps.Streams.Warning.Select(x => x.Message).ToArray();
-            bool hadErrors = w.Ps.HadErrors || errs.Length > 0;
+            bool hadErrors = w.Ps.HadErrors || errs.Length > 0 || cancelled;
 
             // The user script returns a single string from ConvertTo-Json.
             // Embed it as a JSON-string value (escaped).
             string outputJson;
-            if (invokeError != null)
+            if (invokeError != null && !cancelled)
             {
                 WriteError(w.Id, invokeError);
                 return;
             }
-            if (output == null || output.Count == 0 || output[0] == null)
+            if (cancelled || output == null || output.Count == 0 || output[0] == null)
             {
                 outputJson = "null";
             }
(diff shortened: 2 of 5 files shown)


usepowershell:

**Verdict: APPROVE.** This satisfies the cancellation-propagation hard gate I called out on PR #205. Code matches the design doc; per-mode behavior is sound; the Single-mode refactor is the right shape; ProcessPool's existing kill-on-timeout backstop is preserved verbatim (no edits to `OutOfProcessSubprocessPool.cs`).

---

### Design conformance

- Wire protocol matches §3 of the design verbatim: `cancel` request frame with `cancel-` id prefix, `params.requestId` carrying the original invoke id, optional `cancelled` flag on the invoke response, ack frame body shape. Cancel-frame ids are intentionally NOT registered in `_pending`; the read loop downgrades the unknown-id warning for `cancel-` prefixes and for late `cancelled:true` responses to Debug — exactly per §4.2.
- Wire protocol diverges in one harmless place: the design's `cancelled:true` example shows `errors:["The pipeline has been stopped."]`. The implementation passes through whatever `w.Ps.Streams.Error` happens to contain (often empty). Functionally equivalent — the .NET caller has already observed `OperationCanceledException` and `_pending.TryRemove` has fired by the time the response arrives, so the response body is dropped.
- One justified divergence (already noted in Bender's history): `OutOfProcessSubprocessPool.InvokeAsync` did NOT receive the diagnostic `catch (OperationCanceledException)` block from §4.3. OCE bubbles up unannotated. Acceptable — the design called the catch "purely for clean diagnostics," and skipping it keeps the diff minimal. If logs prove confusing later, add it.
- Single-mode also diverges from §5.1 in implementation strategy: instead of `BeginInvoke` + `ThreadPool.QueueUserWorkItem`, the PR compiles a C# `SingleDispatcher` (`BlockingCollection` + dedicated worker thread + `ConcurrentDictionary` registry) that mirrors the existing `PoolDispatcher` shape. **This is a better choice than the design proposed.** Code-share with Pool is high (same `SingleStdout`/`PoolStdout.Lock` pattern, same active-item tracking, same `wasStopped` detection). Avoids fighting PowerShell-only async ergonomics. The `ThreadPool.QueueUserWorkItem` approach the design sketched would have spawned one OS thread per in-flight invoke and put the response-write on a thread-pool worker — fine, but less uniform with Pool.

### Per-mode soundness

**Single (`oop-host.ps1`):** The structural blocker is gone — `Invoke-InvokeHandler` no longer calls `& $cmdInfo @boundParams` synchronously on the dispatcher loop. The handler builds a `[powershell]` instance (assigned to `$script:SharedRunspace`), pre-compiles the user script with `try { & $Name @Splat } catch { throw }` + inline `ConvertTo-Json` fallback chain (covers #203), and submits to `SingleDispatcher`. Dispatcher main loop returns immediately to read the next stdin frame. ✓
- Stdout serialization: `SingleStdout.Lock` synchronizes both worker-thread response writes and `Write-NdjsonResponse` calls (which now route through `[PoshMcp.SingleHost.SingleStdout]::Write`). No interleave possible. ✓
- Error semantics: terminating errors from the user pipeline surface via `PipelineStoppedException` or `Exception` in `ProcessOne` and produce an `error` frame. Non-terminating errors and warnings flow via `w.Ps.Streams.Error/Warning`. The user script's `$Error.Clear()` runs inside the shared runspace per invoke — the leak fix from #189 is preserved. ✓
- Ordering for `wasStopped` detection: catches `PipelineStoppedException` AND falls back to `InvocationStateInfo.State == PSInvocationState.Stopped`. **This belt-and-suspenders is correct** — `BeginStop` does not always raise `PipelineStoppedException` from `Invoke()`; sometimes the synchronous `Invoke()` returns normally with `State=Stopped`. Bender's history note on this is accurate.
- Shared runspace lifecycle: `Ensure-SharedRunspace` lazy-opens once; `Invoke-ShutdownHandler` closes/disposes both dispatcher and runspace. ✓

**Pool (`oop-host-pool.ps1`):** Surgical — `_active` registry added to `PoolDispatcher`, populated in `Submit` before enqueue, removed in worker `finally` after dispose. `Cancel(requestId)` is a `TryGetValue` + `BeginStop`. ProcessOne grew the same `wasStopped` detection. ✓
- **No head-of-line blocking** confirmed: workers iterate `_queue.GetConsumingEnumerable()` independently; cancelling one item only `BeginStop`s its own `[powershell]`. Other runspaces in the pool keep flowing. The Pool test exercises this directly with `runspacePoolSize:4` + parallel slow + fast invoke (fast completes before cancel fires).
- Cancel race-free: if cancel arrives between worker completion and `_active.TryRemove`, `BeginStop` on an already-completed `[powershell]` is a no-op (the dispatcher's `try/catch` swallows any throw). Ack returns `cancelled:true` in this corner — benign.
- Frame ordering under `PoolStdout.Lock`: ack frame from `Invoke-CancelHandler` and the eventual `cancelled:true` response from the worker are both serialized through the same lock. No interleave. ✓

**ProcessPool (`OutOfProcessSubprocessPool`):** Inheritance is genuinely free — the file is **not modified by this PR**. Soft-cancel propagation comes via `OutOfProcessHost.SendRequestAsync` (which Single-mode hosts in the pool call). The per-request kill-on-timeout backstop at line 421 (`catch (TimeoutException tex)`, `MarkSlotDead(slot)`) is preserved verbatim. ✓ The ProcessPool test asserts `pool.HealthyCount >= 1` after soft-cancel — slots stay healthy, kill path not invoked, exactly as designed.

### .NET hook (`SendRequestAsync`)

- Two `Register` callbacks (timeout + caller cancel), each calling `TrySendCancelFrameAsync` (guarded against the `method == "cancel"` recursion case). Both registrations dispose in `finally`. ✓
- Detail worth noting: `timeoutCts` was changed from `CreateLinkedTokenSource(cancellationToken)` to a plain new CTS. Caller cancel and per-request timeout are now properly orthogonal — no double-fire of the timeout path when the caller cancels. The `using` block disposes the timer when the awaiter completes either way. ✓
- `TrySendCancelFrameAsync` uses an independent 2-second CTS for the stdin send so caller-token cancellation cannot poison it. Errors logged at Debug. The defensive `try { _sendLock.Release(); } catch (ObjectDisposedException)` is unusual (SemaphoreSlim doesn't normally throw OOC there) but harmless.
- Cleanup ordering: `cancelRegistration.DisposeAsync` runs after `timeoutRegistration.DisposeAsync` in `finally`; outer `finally` removes `_pending[id]`. No leaks.
- Cancel-races-with-success: if the read loop sets the result a tick before the caller cancels, `tcs.TrySetCanceled` returns false and the awaiter sees the success — but `TrySendCancelFrameAsync` still fires (no completion gate). The host gets a cancel for an already-completed id, replies `cancelled:false`, suppressed at the read loop. Noise-only; not worth gating.

### Tests (`OutOfProcessCancellationTests`)

- One test per mode — minimum required, well-targeted.
- Each test asserts: (a) the awaiter throws `OperationCanceledException` within `ObservationTimeout=15s` against a `Start-Sleep -Seconds 60` (so a passing test means cancel actually unblocked, not that the sleep finished), (b) the host/pool is reusable for a follow-up invoke. Pool adds the head-of-line check (concurrent fast invoke completes during the 500ms warmup window, before cancel fires). ProcessPool adds `HealthyCount >= 1` after a 1s settle.
- 500-750ms warmup before cancel is realistic — enough to ensure the pipeline is mid-`Start-Sleep` rather than still being dispatched. Less would race the invoke send (Bender's history note on this is correct).
- 15s deadline against a 60s sleep gives plenty of headroom. Not tight enough to be flaky; not loose enough to mask a regression.
- Pool test correctly forces `runspacePoolSize:4` via `setup` so the pool is provably > 1 — without this, head-of-line could pass falsely on a 1-runspace pool.
- ProcessPool test's `MinHealthyForStartup:1` + `ReconcilerInterval:500ms` is the right minimum for a 2-slot pool exercising soft cancel.

### Edge cases not tested but worth noting (non-blocking)

- **Cancel arriving before invoke begins:** caller cancels before `Submit` runs in Single, or before `Add` to `_queue` in Pool. Cancel handler's `_active.TryGetValue` returns false → ack `cancelled:false`. The .NET awaiter still sees OCE (registration fires). Benign; no test needed.
- **Cancel arriving on an already-completed request:** covered by the response-suppression path in the .NET read loop and by `cancelled:false` ack. Functionally tested implicitly by the Pool test (the fast `Get-Date` invoke completes before any cancel).
- **Cancel during host startup / shutdown:** `_disposed` and `_stdin is null` checks in `TrySendCancelFrameAsync` guard the host side. The 2s CTS bounds the worst case. Not tested; low risk.
- **Wedged unmanaged-code pipeline ignoring `BeginStop`:** the .NET awaiter still returns OCE promptly (TCS is set independently of host ack). The pipeline keeps running until exit; in ProcessPool, the next invoke on that slot will trigger the existing kill-on-timeout. In Single/Pool, the per-request timeout still fires as backstop. Documented in design §7. Not tested directly; existing `Pool_PerRequestTimeout_KillsHostAndPoolRecovers` covers the kill path for ProcessPool.
- **Vestigial `try { ... } catch { throw }`** in the Single-mode user script is a semantic no-op (a bare `& $Name @Splat` would already propagate terminating exceptions). The comment explains the intent but the wrapper isn't doing what the comment claims. Cosmetic — strip in a follow-up if/when you touch the file.

### CI

`Squad CI / test`, `CodeQL Analyze (actions)`, `Analyze (python)`, dependency submission all green. `CI / build` and `CodeQL Analyze (csharp)` were still in progress at review time; they should pass given the green test job and that this is additive code with no signature changes.

---

### #196 hard gate status — SATISFIED

Both hard gates I called out on PR #205 for the default-mode flip are now closed:

1. ✅ **Custom PSHost/PSHostUserInterface for runspace pool** — landed in PR #201, verified at review time.
2. ✅ **Cancellation propagation** — landed by this PR. In-flight pipelines stop within bounded time across all three modes (Single via shared-runspace `BeginStop`; Pool via per-runspace `BeginStop`; ProcessPool via Single-mode inheritance + existing kill backstop). No head-of-line block in Pool. Hosts/slots stay healthy after soft cancel.

### #196 remaining scope

Now unblocked for the default flip. The list as I see it from spec 004 + this PR:

1. **Default-mode flip:** change `SubprocessHostMode.Default` (or whatever the resolution sentinel is) to `Pool`. Keep `Single` and `ProcessPool` as supported opt-in modes; `ProcessPool` remains the recommended choice for tail-sensitive / isolation-sensitive workloads (per #195 findings: P99 within 0.7ms of mean).
2. **Config key naming review:** confirm the `appsettings.json` surface for `SubprocessHostMode` reads naturally (`PowerShellConfiguration.Execution.OutOfProcessMode` or similar). Confirm enum-vs-string serialization story is consistent with how the rest of the config surface handles enums. Reconfirm there's no lingering `SubprocessHostMode` collision (Bender's enum vs Hermes's static-class merge order from PRs #200/#201 — should already be resolved by this point but worth a final audit).
3. **Doctor validation hooks:** doctor should surface (a) the resolved `OutOfProcessMode`, (b) for Pool, the resolved `RunspacePoolSize` and any clamp applied (recall #201's hardcoded cap of 8), (c) the host script path that was resolved, (d) the per-request timeout. Failure modes: warn (don't error) if Pool is configured but pwsh resolution failed; doctor should still succeed against in-process fallback.
4. **Doc updates:** README, DOCKER.md, spec 004 supersedence note. State the new default. State that ProcessPool is the opt-in for isolation/tail latency. Document the cancellation contract (caller-token cancel → bounded soft-cancel; per-request timeout as backstop; ProcessPool kill-on-timeout backstop preserved).
5. **Bench reaffirmation (per my #205 position):** Hermes runs a `--job long` `WarmInvokeThroughputBenchmark` against post-this-PR `main` (run-4) and confirms the ≥ 4× I/O bar still holds. The cancellation refactor introduces additional per-invoke object allocation (the `[powershell]` instance + dispatcher hop). I expect no measurable regression on warm I/O — but verify, don't assume.

I do not see any new gates from reading #207. The two I named on #205 are the complete set.

PR #222, opened 2026-05-12
fix: SwitchParameter round-trips through MCP JSON arguments

youyuanwu:
PowerShell SwitchParameter is a struct with a getter-only IsPresent property. The MCP SDK's reflection-based JSON binding reads it as a default(SwitchParameter) regardless of the wire payload, so every [switch] cmdlet parameter silently arrived as IsPresent=false. The same reflection produces a schema of {"type":["object","null"],"properties": {"isPresent":{"type":"boolean"}}}, which most MCP clients reject when the model emits a natural boolean.

This change introduces SwitchParameterJsonConverter + SwitchParameterMcpSupport (single file under PoshMcp.Server/PowerShell) and wires both into every generated MCP tool via
McpServerToolCreateOptions.SerializerOptions and SchemaCreateOptions:

* Converter accepts boolean / null / numeric / string / {isPresent} / empty object on read, emits plain boolean on write.
* TransformSchemaNode rewrites the SwitchParameter schema node into anyOf [boolean | {isPresent} | null] so natural payloads validate while older clients that emit the struct envelope still work.

Tests (PoshMcp.Tests/Functional/SwitchParameterMcpRoundTripTests.cs):
* 12 converter cases covering every payload shape.
* Regression guard documenting the silent-false bug if the converter is ever removed.
* Schema assertions on a real Get-ChildItem -Recurse tool — verifies the anyOf is emitted and that non-switch parameters (-Path) are untouched.
* End-to-end: define a PowerShell function with a [switch], generate the MCP method, invoke it through CreateParameterArray + method.Invoke with each wire JSON shape, and assert the function actually saw IsPresent={true,false} as appropriate.

### DIFF
--- a/PoshMcp.Server/McpToolFactoryV2.cs
+++ b/PoshMcp.Server/McpToolFactoryV2.cs
@@ -557,7 +557,15 @@ public class McpToolFactoryV2
             OpenWorld = true,
             ReadOnly = metadata.IsReadOnly,
             Title = metadata.CommandName,
-            UseStructuredContent = true
+            UseStructuredContent = true,
+            // Carries SwitchParameterJsonConverter so JSON like {"isPresent": true}
+            // (or a plain bool, if the schema permits it) actually round-trips back
+            // to a real SwitchParameter(true) at invocation time. Without this, every
+            // PowerShell switch parameter silently arrived as IsPresent=false.
+            SerializerOptions = SwitchParameterMcpSupport.SerializerOptions,
+            // Rewrites the SwitchParameter schema node into an anyOf that accepts
+            // plain boolean, the legacy {isPresent} envelope, and null.
+            SchemaCreateOptions = SwitchParameterMcpSupport.SchemaOptions,
         };
     }
 
--- a/PoshMcp.Server/PowerShell/PowerShellParameterUtils.cs
+++ b/PoshMcp.Server/PowerShell/PowerShellParameterUtils.cs
@@ -299,6 +299,14 @@ public static class PowerShellParameterUtils
     /// (boxed string); the proxy function happily accepts it and forwards it
     /// over the implicit-remoting channel where the real underlying cmdlet
     /// performs its normal type coercion (string → Uri / enum / int / etc.).
+    /// <para>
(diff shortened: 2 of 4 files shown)


usepowershell:
## Approval — Hermes (PowerShell Engineer)

Rebased onto current `main` (`c70259c`) cleanly — no conflicts. Build clean (0 errors). Test results from the rebased branch:

- **Unit:** 532 / 532 ✅
- **Functional:** 125 passed, 1 skipped (unrelated), 0 failed ✅ — including the new 21 `SwitchParameter*` cases (12 converter shapes + serializer round-trip + bare-STJ regression guard + 2 schema assertions + 5 e2e PowerShell invocations)

### Code review

**Correctness / wiring** — Converter + schema transform are both registered on `McpServerToolCreateOptions` inside `BuildToolCreateOptions`, which is the central path for every generated tool. All tools get the fix, not a subset.

**JSON deserialization safety** — `Read` defends every `JsonTokenType`: bool/null direct, string via `bool.TryParse`, number via `TryGetInt64`, object loop uses `reader.Skip()` for unknown properties (no recursive overhead), default falls through to `false`. STJ's built-in 64-level nesting guard handles the DoS surface; the converter doesn't introduce any unbounded allocation or recursion of its own. Empty-object => presence is intentional and documented.

**Schema correctness** — `anyOf [boolean | {object: isPresent} | null]` is valid JSON Schema and matches what MCP clients in the wild actually emit. Backwards compat is preserved (legacy `{isPresent: true}` envelope still validates). Verified by `GeneratedToolSchema_LeavesNonSwitchParametersUntouched` that `-Path` (`string[]`) is not rewritten — the transform is correctly type-gated.

**Spec 010 interaction** — Orthogonal layers. Spec 010 added `IToolMetadataSource` / description-source tracking / OTel counters, all in the *description resolution* path. This PR adds `SerializerOptions` + `SchemaCreateOptions` to the same `McpServerToolCreateOptions` initializer. Rebase needed no manual conflict resolution. Both sets of changes preserved.

**Test claims match coverage** — All names accurately describe what the test does. The regression guard explicitly documents the bare-STJ failure mode so a future reviewer can't silently drop the converter.

Approving and merging via squash.

Issue #228, opened 2026-05-12
OOP: wire path through the shared IToolMetadataSource


usepowershell:
Per spec 010 sequencing step 6.

Route OOP description resolution through the same `IToolMetadataSource` as the in-process path, sourcing input from the new `RemoteToolSchema` fields. Remove the OOP-specific Synopsis-only fallback in `oop-host.ps1#L763-L771` and `oop-host-pool.ps1#L824-L832` once the schema fields carry the same data the in-process path consumes.

**Spec:** [specs/010-tool-self-documentation/spec.md](../tree/main/specs/010-tool-self-documentation/spec.md) (FR-520, Sequencing 6)

**Acceptance:**
- Same code produces the same descriptions in both modes (covered by parity test issue)

PR #238, opened 2026-05-12
feat(metadata): extract IToolMetadataSource seam (#225)

usepowershell:
Closes #225

Introduces IToolMetadataSource per spec 010 Option A. Seam only - no behavior change. Foundational for #226, #227, #228.

## What's in this PR

- PoshMcp.Server/PowerShell/IToolMetadataSource.cs - interface + request/result records, source enums (values align 1:1 with spec 010 FR-583 descriptionSource literals).
- PoshMcp.Server/PowerShell/DefaultToolMetadataSource.cs - pass-through implementation preserving pre-spec-010 behavior byte-for-byte.
- McpToolFactoryV2 - both SetParameterSetDescription (in-process) and CreateRemoteCommandMetadataMapping (OOP) route through the seam.
- StdioServerHost / HttpServerHost - TryAddSingleton<IToolMetadataSource, DefaultToolMetadataSource>() so #226/#227 can swap in a replacement via DI.
- McpToolSetupService.CreateToolFactory - accepts optional IToolMetadataSource, defaults to no-op.

## Behavior preservation

Default impl reproduces existing paths: in-process Syntax->Name fallback identical; OOP Synopsis-when-non-empty identical to schema.Description path.

## Not in this PR (deferred)

- Get-Help precedence chain (FR-500/FR-510): #226 + #227.
- RemoteToolSchema wire-format extension: #227.
- Doctor descriptionSource + metrics counters: #228.
- PowerShellSchemaGenerator parameter-description wiring: deferred to #226.

## Validation

- dotnet build PoshMcp.sln -c Release: 0 errors. 20 warnings all pre-existing.
- dotnet test --filter Category!=Integration: 661 passed, 7 skipped, 0 failed.

## Spec gaps

None material. Minor reviewer call: ToolDescriptionRequest.LongDescription is on the interface but default impl ignores it (spec assigns long-description sourcing to #226's caller-side Get-Help integration, not the seam itself).

Reviewers: Farnsworth + Cubert (squad).

PR #236, opened 2026-05-12
test: pre-spec010 tools/list snapshots (#224)

usepowershell:
Closes #224

Captures baseline tools/list JSON for in-process and OOP modes per FR-550 (spec 010 step 2).

**What this PR adds**
- specs/010-tool-self-documentation/baseline/inprocess-tools-list.json (133 tools)
- specs/010-tool-self-documentation/baseline/oop-tools-list.json (144 tools)
- specs/010-tool-self-documentation/baseline/capture-snapshots.ps1 — reproducible capture script
- specs/010-tool-self-documentation/baseline/README.md — capture date, server SHA, commands, regeneration policy
- PoshMcp.Tests/Fixtures/Modules/HelpParityFixture/ — the FR-521 fixture module (six deterministic functions, one per precedence-chain rung). Required to capture FR-500/FR-510 baselines and reused later by the parity + regression tests.

**Capture protocol**
The script launches dotnet PoshMcp.dll serve --transport stdio per mode, speaks initialize → notifications/initialized → tools/list, and persists the full JSON-RPC envelope pretty-printed (2-space indent, LF endings) for diff-friendly review.

**Notable pre-change parity artifacts (already documented in baseline/README.md)**
- Tool-count delta (133 vs 144) — the two runtime paths expose slightly different command surfaces today for the same configured module set. Out of scope for spec 010 (FR-551).
- The in-process discovery path requires explicit `CommandNames` to auto-load modules from PSModulePath; the OOP path imports `config.Modules` during `setup`. Captured as-is.
- `Environment.ImportModules` / `Environment.ModulePaths` are honored only by the OOP path in the current code (`PowerShellEnvironmentSetup` is unwired for in-process). The capture works around it by setting `PSModulePath` on the spawned process.

Reviewers: Farnsworth + Cubert (squad).

PR #235, opened 2026-05-12
docs: fix misleading RemoteToolSchema XML doc (#233)

usepowershell:
Closes #233

Corrects XML doc on RemoteToolSchema.Description to match current OOP behavior per FR-560 (spec 010 step 10). Doc-only change, no runtime impact.

The previous text claimed the field came from `Get-Help` *or* parameter set syntax. In reality, the OOP host (`oop-host-pool.ps1` ~L824-829) only reads `Get-Help .Synopsis`, trims it, and assigns it when it differs from the command name; otherwise the field is the empty string (which downstream becomes the bare command name).

No other stale property docs spotted in this file.

Reviewers: Farnsworth + Cubert (squad).

PR #239, opened 2026-05-13
feat(oop): extend RemoteToolSchema with full description + per-parameter metadata (#227)

### LINKED ISSUE #227: OOP: extend RemoteToolSchema with full description and per-parameter description
Per spec 010 sequencing step 5.

Extend `PoshMcp.Server/PowerShell/OutOfProcess/RemoteToolSchema.cs` and the host scripts (`oop-host.ps1`, `oop-host-pool.ps1`) to emit the additional metadata fields the precedence chain needs: full Description body, per-parameter description, ParameterAttribute.HelpMessage, ValidateSet values. The existing `Description` field MUST remain populated for backward compatibility.

**Spec:** [specs/010-tool-self-documentation/spec.md](../tree/main/specs/010-tool-self-documentation/spec.md) (FR-500, FR-510, Sequencing 5)

**Acceptance:**
- Schema extension is additive (no breaking field changes)
- Existing `Description` populated as before for older clients

usepowershell:
Closes #227

Adds additive metadata fields to RemoteToolSchema and emits them from oop-host.ps1 + oop-host-pool.ps1 per FR-500/FR-510. Existing Description field unchanged for backward compatibility. Consumer side wiring is in #228.

Reviewers: Farnsworth + Cubert (squad).

### DIFF
--- a/PoshMcp.Server/PowerShell/OutOfProcess/RemoteToolSchema.cs
+++ b/PoshMcp.Server/PowerShell/OutOfProcess/RemoteToolSchema.cs
@@ -17,11 +17,36 @@ public class RemoteToolSchema
     /// Human-readable description for the command. Populated by the OOP host
     /// from <c>Get-Help</c>'s <c>.Synopsis</c> field (trimmed, and only when
     /// it differs from the command name); otherwise the empty string. The
-    /// long <c>Description</c> body and parameter set syntax are NOT read.
+    /// long <c>Description</c> body is exposed separately via
+    /// <see cref="FullDescription"/>; parameter set syntax is NOT read here.
     /// When empty, downstream tool schema generation falls back to the bare
     /// command name.
     /// </summary>
+    /// <remarks>
+    /// Preserved verbatim for backward compatibility with consumers that
+    /// only know about this field. Spec 010 introduces <see cref="FullDescription"/>
+    /// alongside it; the precedence chain (FR-500) is applied on the .NET
+    /// consumer side, not by the host script.
+    /// </remarks>
     public string Description { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Optional long-form description body sourced from
+    /// <c>Get-Help &lt;command&gt;.Description</c>, an array of
+    /// <c>MamlParaText</c> objects whose <c>.Text</c> values are joined with
+    /// the paragraph separator <c>"\n\n"</c>. Raw text — no sanitization or
+    /// length capping is applied by the host; the .NET consumer applies
+    /// FR-540 sanitization and FR-541 length caps.
+    /// </summary>
+    /// <remarks>
+    /// <c>null</c> when <c>Get-Help</c> returned no help record, threw, or
+    /// returned a record whose <c>.Description</c> property was absent or
+    /// empty. The empty string is also possible when <c>Description</c>
+    /// existed but contained only whitespace after the join.
+    /// Consumers MUST treat <c>null</c> and empty-or-whitespace as
+    /// "no value" and fall through to the next precedence step.
+    /// </remarks>
+    public string? FullDescription { get; set; }
 
     /// <summary>
     /// The parameter set name this schema represents
@@ -52,4 +77,48 @@ public class RemoteParameterSchema
 
     public bool IsMandatory { get; set; }
     public int Position { get; set; } = int.MaxValue;
+
+    /// <summary>
+    /// Optional per-parameter description text sourced from
+    /// <c>Get-Help &lt;command&gt;.Parameters.parameter[name=&lt;Name&gt;].description</c>,
+    /// joined paragraph-by-paragraph with <c>"\n\n"</c>. Raw text — the
+    /// .NET consumer applies FR-540 sanitization and FR-542 length caps.
+    /// </summary>
+    /// <remarks>
+    /// <c>null</c> when the help record had no entry for this parameter, the
+    /// entry's <c>description</c> array was absent, or <c>Get-Help</c>
+    /// itself was unavailable. Consumers treat <c>null</c> or
+    /// empty-or-whitespace as "no value" and fall through to the next
+    /// precedence step (HelpMessage → ValidateSet → type fallback) per
+    /// FR-510.
+    /// </remarks>
+    public string? HelpDescription { get; set; }
+
+    /// <summary>
+    /// Optional <c>HelpMessage</c> value from <c>[Parameter(HelpMessage="...")]</c>
+    /// on this parameter, read from <c>CommandInfo.Parameters[Name].Attributes</c>
+    /// of type <c>System.Management.Automation.ParameterAttribute</c>.
+    /// </summary>
+    /// <remarks>
+    /// <c>null</c> when no <c>ParameterAttribute</c> sets <c>HelpMessage</c>,
+    /// or when the attribute was unavailable. If multiple parameter sets
+    /// declare the parameter with different <c>HelpMessage</c> values, the
+    /// first non-empty value encountered is emitted; the host does not
+    /// attempt to reconcile divergent values across attributes.
+    /// </remarks>
+    public string? HelpMessage { get; set; }
+
+    /// <summary>
+    /// Optional list of allowed values from <c>[ValidateSet(...)]</c> on this
+    /// parameter, read from the parameter's
+    /// <c>System.Management.Automation.ValidateSetAttribute.ValidValues</c>
+    /// collection.
+    /// </summary>
+    /// <remarks>
+    /// <c>null</c> when the parameter has no <c>ValidateSetAttribute</c>.
+    /// An empty array is theoretically possible but treated equivalently to
+    /// <c>null</c> by FR-510 step 3. Order is preserved as declared in the
+    /// attribute.
+    /// </remarks>
+    public string[]? ValidateSetValues { get; set; }
 }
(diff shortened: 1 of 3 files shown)


usepowershell:
## Architecture review (Lead) — PR #239

**Verdict: would approve.** Strictly additive wire-format extension, defensive PowerShell on both host scripts, exact C#/JSON shape parity. No architectural blockers; notes below are non-blocking observations.

### What was checked

- **Backward compatibility (additive).**
  - `RemoteToolSchema.Description` is preserved verbatim — same field name, same `string` type, same default (`""`), same population semantics in both host scripts (Get-Help `.Synopsis`, trimmed, suppressed when equal to command name). XML doc was updated to point at `FullDescription` for the long body but the field's runtime contract did not change.
  - `RemoteToolSchema.FullDescription` is `string?` — older serialized payloads (pre-spec-010 hosts) deserialize with `null`, which the consumer (per FR-510 doc) treats as "no value." Same for `RemoteParameterSchema.HelpDescription`, `HelpMessage`, and `ValidateSetValues` (all nullable).
  - All four new fields are properties with parameterless getters/setters — the standard `System.Text.Json` contract preserves backward read/write.

- **JSON ↔ C# shape parity.** Spot-checked the ordered hashtables emitted by both host scripts against the DTO:

  | Layer | Host script keys | C# DTO properties | Match |
  |---|---|---|---|
  | Tool | `Name`, `Description`, `FullDescription`, `ParameterSetName`, `Parameters` | `Name`, `Description`, `FullDescription`, `ParameterSetName`, `Parameters` | ✓ |
  | Parameter | `Name`, `TypeName`, `IsMandatory`, `Position`, `HelpDescription`, `HelpMessage`, `ValidateSetValues` | `Name`, `TypeName`, `IsMandatory`, `Position`, `HelpDescription`, `HelpMessage`, `ValidateSetValues` | ✓ |

  Both `oop-host.ps1` (Single) and `oop-host-pool.ps1` (Pool) emit the same shape; the .NET consumer needs no per-host-mode branching.

- **Defensive PowerShell — null-safe at every layer.**
  - Outer `Get-Help` call: `try { Get-Help -Name $cmd.Name -ErrorAction SilentlyContinue } catch { $helpInfo = $null }`. Discovery cannot fail because help is missing (FR-581).
  - `Get-RemoteCommandHelpMetadata` (single host) / `Get-HelpMeta` (pool host): every property read is guarded by `if ($null -ne $HelpInfo)` and inner reads are wrapped in `try { ... } catch { }`. Returns sentinel object with `Synopsis = ''`, `FullDescription = $null`, empty `ParamHelp` hashtable on any failure.
  - MAML projection: each `description` member walked element-by-element; both string-array and `MamlParaText[]`-array shapes handled (`if ($d -is [string]) { $t = $d } elseif ($null -ne $d -and $null -ne $d.Text) { $t = "$($d.Text)" }`). Skips `$null` entries cleanly.
  - `Get-RemoteParameterAttributeMetadata` / `Get-ParamAttrMeta`: full try/catch around the attribute walk; checks `Cmd.Parameters` and `pmeta.Attributes` for `$null` before iterating; checks each attribute for `$null`. `HelpMessage` only set when the string is non-empty after `[string]::IsNullOrWhiteSpace` check. `ValidateSetValues` only emitted when the array has at least one entry (matches the C# `null` ↔ "no value" contract).

- **Coverage of the OOP host gap raised in spec 010 wave 1.** Both `oop-host.ps1` and `oop-host-pool.ps1` are updated in this PR. The asymmetry (script-scope helpers in `oop-host.ps1`, nested local functions in `oop-host-pool.ps1`'s `Invoke-DiscoverHandler`) is intentional and matches the established pool-host pattern from PR #204 / #207 — pooled runspace scriptblocks should not depend on host-process functions.

- **No precedence resolution in the host.** `Description` continues to be populated only with `Synopsis` (sanitized to drop the auto-generated case) — the host does not pre-resolve to the long Description body or the syntax fallback, which would race the spec 010 precedence chain on the consumer side. The `FullDescription` field carries raw paragraph-joined text, FR-540 sanitization happens on the .NET consumer per the seam contract from #238. This is the right separation; the host emits source data only.

### Non-blocking observations

1. **Description-sourcing logic duplicated across two host scripts.** `Get-RemoteCommandHelpMetadata`/`Get-HelpMeta` and `Get-RemoteParameterAttributeMetadata`/`Get-ParamAttrMeta` are functionally identical between `oop-host.ps1` and `oop-host-pool.ps1`. Carrying two copies is the historical pattern (per #204 fallback, #207 dispatcher) and is correct for runspace-isolation reasons in the pool. A future consolidation into a shared `oop-host-shared.ps1` `dot-source`d by both is a candidate refactor but explicitly out of scope here.

2. **Single mode discovery snapshot.** `specs/010-tool-self-documentation/baseline/oop-tools-list.json` (#236) was captured against the Pool host. With this PR the Single host emits the same shape too — capturing an additional snapshot for Single in #229's parity work would close the OOP coverage gap I flagged on #236. Not blocking #239 — flag for #229.

3. **`HelpMessage` "first non-empty" rule across multiple parameter sets.** When the same parameter appears in multiple parameter sets with different `ParameterAttribute.HelpMessage` values, the host emits the first non-empty one encountered (XML doc on `HelpMessage` calls this out explicitly). FR-511 requires per-parameter consistency across parameter sets within a command — the host's "first wins" matches that intent (a single value per parameter), but the doc note is the right place to flag the edge case. No code change needed.

4. **`ValidateSetAttribute.ValidValues` ordering.** Host preserves declaration order via `foreach ($v in $attr.ValidValues) { $vals += "$v" }`. FR-510 step 3 specifies declaration order — matches.

5. **Consumer wiring lands in #228.** Confirmed: `McpToolFactoryV2.CreateRemoteCommandMetadataMapping` is unchanged in this PR. After #240 (in-process precedence) and #239 (OOP schema extension) both land, #228 wires the OOP consumer through `IToolMetadataSource` using the new `RemoteToolSchema` fields, at which point FR-520 byte-parity becomes verifiable.

### XML doc precedence (FR-560)

`RemoteToolSchema.Description` XML doc was already corrected in #235 for the pre-spec-010 contract. This PR's update to that doc (mentioning `FullDescription`) is accurate for the new wire format but still describes "Synopsis only" semantics for the `Description` field itself — that remains accurate as long as #228 routes the long-form text through `FullDescription`. When #228 lands and consumers source descriptions through the precedence chain, `Description`'s XML doc should be updated again to describe its role as one input among several to the FR-500 chain. Flag for #228.

`gh pr review` remains EMU-blocked from this account; this is a comment, not a formal GitHub Approve.

usepowershell:
## 🔍 Cubert (Fact Checker) verification — PR #239

**Confidence:** ✅ HIGH — APPROVE.

### What I checked
1. `gh pr view 239` / `gh pr diff 239` (+329 / -25, 3 files; CI all green; mergeable)
2. Read every changed file in worktree `poshmcp-227` (read-only)
3. **Additive schema:** confirmed `RemoteToolSchema.Description` is unchanged (`public string Description { get; set; } = string.Empty;`) and every new field is **nullable with null default** — `FullDescription`, `RemoteParameterSchema.HelpDescription`, `HelpMessage`, `ValidateSetValues`. XML-doc on each new field explicitly notes the C# consumer applies FR-540/541/542 (host emits raw source data only — correct separation).
4. **PS host emission parity:** read both diffs in detail — `oop-host.ps1` (top-level helpers `Get-RemoteCommandHelpMetadata`, `Get-RemoteParameterAttributeMetadata`) and `oop-host-pool.ps1` (inline `Get-HelpMeta`, `Get-ParamAttrMeta`). Both emit identical PascalCase keys: `Name`, `Description`, `FullDescription`, `ParameterSetName`, `Parameters`, `HelpDescription`, `HelpMessage`, `ValidateSetValues`, `TypeName`, `IsMandatory`, `Position` — matching the C# DTO property names exactly.
5. **JSON deserialization:** confirmed `OutOfProcessCommandExecutor.cs:124` uses `var options = new JsonSerializerOptions { PropertyNameCaseInsensitive = true };` for the discover-response deserialization. PascalCase keys from PS map cleanly; older/missing optional fields are tolerated by `System.Text.Json` defaults.
6. **`Description` field semantics preserved:** the synopsis predicate is unchanged — `Description = $helpMeta.Synopsis` where `Get-RemoteCommandHelpMetadata` only sets `Synopsis` when it is non-empty AND `!= command name`. Same logic as before — synopsis if real, empty string otherwise.
7. **Empirical backward-compat test:** built the worktree (`Release`), re-ran `specs/010-tool-self-documentation/baseline/capture-snapshots.ps1`, and diffed the freshly-captured OOP snapshot against the committed baseline (`oop-tools-list.json`). Result:

```
BaseTools: 144  NewTools: 144
Tool description diffs: 0
Missing tools: 0
```

**Every single one of the 144 OOP tools has a byte-identical `description` field.** Restored the baseline files afterwards; nothing committed or pushed.

### Verdict
✅ **APPROVE.** The schema change is purely additive: existing `Description` field semantics are preserved verbatim, all new fields are nullable with null defaults, both PS hosts emit matching PascalCase keys, and `PropertyNameCaseInsensitive=true` makes the deserialization safe in either direction. An older client deserializing the new JSON against the old DTO simply ignores the extra fields (S.T.J default behavior). Empirical proof: zero description diffs across the entire 144-tool surface.

> ⚠️ Posting via `gh pr comment` instead of `gh pr review --approve` because of the EMU/usepowershell self-review block. This comment does **not** count as a formal GitHub approval — please get a project-maintainer approval before merge.

— Cubert (Fact Checker), wave 3 review 2026-05-12, requested by Steven

Issue #242, opened 2026-05-13
FR-510 wiring gap: inputSchema parameter descriptions are empty in both InProcess and OOP modes


usepowershell:
## Bug

`inputSchema.properties.<name>.description` is **empty for every parameter on every tool** in both InProcess and OutOfProcess modes, even for commands whose parameters have authored `.PARAMETER` help blocks, `[Parameter(HelpMessage="...")]` attributes, or `[ValidateSet(...)]` constraints.

This was first observed by Cubert during PR #241 verification ("Parameter-level `inputSchema.properties.<name>.description` is empty in both modes for every tool in both snapshots (0 populated of all parameters)"). The FR-520 parity assertion in #229 passes trivially because both runtime paths emit the same empty string.

## Why this matters

Spec 010 FR-510 defines a four-step precedence chain that should populate parameter descriptions:

1. Get-Help `.Parameters.parameter[].description`
2. `[Parameter(HelpMessage="...")]`
3. `[ValidateSet(...)]` allowed-values phrasing ("One of: …" / "Each item is one of: …")
4. Type fallback ("Parameter of type `<TypeName>`")

The resolver (`HelpAwareToolMetadataSource.ResolveParameterDescription`) implements this correctly — its unit tests in `ParameterSetConsistencyTests.cs` all pass — and the call sites in `McpToolFactoryV2` invoke the resolver. The gap is that the resolved descriptions never reach the generated `inputSchema.properties.<name>.description` JSON output that the MCP SDK serializes for clients.

## Failing tests pinned to this issue

`PoshMcp.Tests/Integration/ToolDescriptionParityTests.cs` contains 10 `[Fact(Skip="Tracking issue #XXX")]` tests pointing at this issue:

- `ParameterDescription_IsNonEmpty_WhenHelpTextAvailable_InProcessMode` (5 inline-data variants)
- `ParameterDescription_IsNonEmpty_WhenHelpTextAvailable_OutOfProcessMode` (5 inline-data variants)

When the wiring gap is fixed, **un-skip these tests** — they are the regression gate for the fix.

## Reproduction

`HelpParityFixture` exists exactly for this: 6 commands designed to exercise each FR-510 step. After running `tools/list` against the fixture, every tool variant has `inputSchema.properties.<name>.description` empty for `Message`, `Count`, `UserId`, `Color`, `Directions` — parameters that all have authored help in `HelpParityFixture.psm1`.

## Suspected root cause

[...]

PR #241, opened 2026-05-13
feat(oop): wire OOP through IToolMetadataSource seam (#228)

usepowershell:
Closes #228

Routes OOP description resolution through the same `IToolMetadataSource` seam used by the in-process path, consuming the new `RemoteToolSchema` fields from #227. Establishes mode parity for tool/parameter descriptions per FR-520. Parity regression tests land in #229.

**Changes**

- `McpToolFactoryV2`
  - `CreateRemoteCommandMetadataMapping` now passes `schema.FullDescription` as `LongDescription` and a synthesized parameter-set syntax string as `ParameterSetSyntax`, so the OOP tool-description chain exercises FR-500 steps 1→4 identically to in-process (`SetParameterSetDescription`).
  - New `BuildRemoteParameterDescriptionMap` mirrors the in-process `BuildParameterDescriptionMap`: it iterates `RemoteParameterSchema` entries and resolves each through `_toolMetadataSource.ResolveParameterDescription` using the new `HelpDescription` / `HelpMessage` / `ValidateSetValues` fields. Output feeds the OOP assembly generator.
  - Helper `BuildRemoteParameterSetSyntax` synthesizes `[-Param <Type>]` / `-Param <Type>` from schema parameters so FR-500 step 3 works without round-tripping `CommandParameterSetInfo.ToString()` over the wire.
- `OutOfProcessToolAssemblyGenerator`
  - New `GenerateAssembly(schemas, parameterDescriptions, logger)` overload (legacy single-arg overload preserved). Per-command description map is forwarded to `GenerateMethodForSchema` and used to attach `[Description]` to each PowerShell-derived parameter — same pattern as `PowerShellAssemblyGenerator` so the MCP SDK auto-schema surfaces them as `inputSchema.properties.<name>.description`.
  - Framework parameters (`_AllProperties`, `_MaxResults`, `_RequestedProperties`) and `CancellationToken` are skipped, matching in-process.

No host-script changes: `oop-host.ps1` already emits the new fields raw per #239; the C# consumer now no longer depends on `Description` (Synopsis) being non-empty since the precedence chain handles fall-through.

**Verification**

- `dotnet build PoshMcp.sln -c Release` — clean (no new warnings).
- `dotnet test PoshMcp.Tests/PoshMcp.Tests.csproj -c Release --filter "Category!=Integration"` — 684 passed, 7 skipped, 0 failed.
[...]

PR #243, opened 2026-05-13
test(parity): tool description parity + regression + param-set consistency tests (#229)

### LINKED ISSUE #229: Add FR-521 parity test + FR-550 regression snapshot test + SC-208 parameter-set consistency test
Per spec 010 sequencing step 7.

Land the three required automated tests in `PoshMcp.Tests`:

1. **Parity test (FR-521):** `PoshMcp.Tests/Integration/ToolDescriptionParityTests.cs`. Fixture corpus at `PoshMcp.Tests/Fixtures/Modules/HelpParityFixture/HelpParityFixture.psm1` with 5 named functions covering each precedence step. Equality scoped to `description` + `inputSchema.properties.<name>.description`. Run both modes in a single test session. Pre-warm Get-Help to bound flake.
2. **Regression snapshot test (FR-550):** Compare post-change `tools/list` against the baseline captured under `specs/010-tool-self-documentation/baseline/`. Assert equal-or-superset for non-empty originals sourced from `.Synopsis`.
3. **Per-parameter-set consistency test (SC-208 / FR-511):** Same parameter MUST surface the same description across parameter sets.

**Spec:** [specs/010-tool-self-documentation/spec.md](../tree/main/specs/010-tool-self-documentation/spec.md) (FR-521, FR-550, FR-511, SC-208)

**Acceptance:**
- All three tests on main, green, in both modes

usepowershell:
Closes #229.

Adds the spec-010 wave-5 test trio so the description-source contract has automated coverage in both runtime modes plus a backward-compatibility gate against the committed pre-spec-010 baselines.

## What's tested

| File | Spec | Coverage |
|------|------|----------|
| `PoshMcp.Tests/Integration/HelpParityFixtureSession.cs` | infra | xUnit fixture: brings up an `InProcessMcpServer` + `ExternalMcpClient` against the `HelpParityFixture` corpus for a chosen `RuntimeMode` (InProcess or OutOfProcess), captures one `tools/list` snapshot, tears down. Reused by both integration test classes. |
| `PoshMcp.Tests/Integration/ToolDescriptionParityTests.cs` | **FR-520** | 13 facts/theories across 3 dimensions: (1) fixture-command-count parity between modes, (2) tool-level description byte-equality across modes for each of the 6 fixture commands, (3) per-parameter description byte-equality across modes for each variant. |
| `PoshMcp.Tests/Integration/ToolDescriptionRegressionTests.cs` | **FR-550** | 2 facts (one per mode): asserts every Synopsis-sourced tool description in the committed `specs/010-tool-self-documentation/baseline/{inprocess,oop}-tools-list.json` snapshots is preserved as equal-or-superset (synopsis can grow into synopsis + paragraph; cannot shrink, change, or be replaced). Heuristic origin check uses live `Get-Help .Synopsis` for fixture commands; non-fixture descriptions are conservatively excluded per the FR-550 carve-out for syntax/bare-name fallbacks. |
| `PoshMcp.Tests/Unit/ParameterSetConsistencyTests.cs` | **SC-208 / FR-511** | 8 unit facts proving `HelpAwareToolMetadataSource` is invariant across (a) repeated invocations with identical inputs, (b) varying `CommandName` while parameter inputs are held constant. Last property is the necessary condition for "applied per parameter, not per parameter set." Also includes a tool-level FR-501 invariance check across `ParameterSetSyntax` variation. |

## Cubert PR #241 finding — confirmed real, gated by #242

PR #241's verification noted that `inputSchema.properties.<name>.description` was empty (0/all populated) for every parameter on every tool in both modes. The new tests confirm this end-to-end against the `HelpParityFixture` corpus, which exists exactly so this gap can be observed:

- 5 in-process + 5 OOP `ParameterDescription_IsNonEmpty_WhenHelpTextAvailable_*` inline-data variants all fail when un-skipped — every parameter description is empty even for `Get-FixtureFullHelp.Message` (.PARAMETER block), `Get-FixtureHelpMessageOnly.UserId` (HelpMessage attr), `Get-FixtureValidateSetScalar.Color` (singleton ValidateSet), `Get-FixtureValidateSetArray.Directions` (array ValidateSet).
- The resolver itself (`HelpAwareToolMetadataSource.ResolveParameterDescription`) returns the right strings — `ParameterSetConsistencyTests` proves this with 8 passing facts. The gap is between the resolver output and the generated MCP `inputSchema` JSON.
- Filed as **#242** ("FR-510 wiring gap: inputSchema parameter descriptions are empty in both InProcess and OOP modes"). The 10 non-empty assertions are converted to `[Theory(Skip="Tracking issue #242 — ...")]` so they un-skip green when the fix lands and serve as the regression gate. The `Skip` message points reviewers at #242.
- FR-520 parity passes today because *both* modes are equally empty. This PR's parity assertions still have value: they catch any future regression where one mode populates and the other doesn't. The FR-510 step in the chain is the bug, not the parity rule.

## Test results

- **New tests in isolation:** `Failed: 0, Passed: 23, Skipped: 2, Total: 25` (the 2 skipped are the `[Theory(Skip)]` methods pinned to #242).
- **Full unit suite (regression check):** `Failed: 0, Passed: 501, Skipped: 0, Total: 501`.
- Build clean, no new warnings.

## Reviewers

Cubert + Farnsworth (squad). Cubert's PR #241 review surfaced the wiring gap; this PR turns the observation into an automated, file-an-issue-shaped regression gate.


usepowershell:

**Verdict: ✅ APPROVE** (one non-blocking nit on the FR-550 superset rule)

### Architecture
- **Separation of concerns is clean.** `HelpParityFixtureSession` is a single, internal, reusable in-proc/OOP harness; the three test classes consume it without bleeding setup logic into each other. Fixture vs parity vs regression vs consistency are four distinct files with one job each — no overlapping assertions.
- **Test categorization is correct.** `ParameterSetConsistencyTests` lives in `Unit/` because `HelpAwareToolMetadataSource` is a pure function — no server, no runspace, no IO. The other three live in `Integration/` because they require `InProcessMcpServer` + `ExternalMcpClient` + a real PowerShell runspace. The `[Trait("Spec","010")]` annotation makes spec-scoped filtering trivial.
- **Lifecycle is sound.** `IAsyncLifetime` boots both `RuntimeMode` sessions once per test class, the OOP session is conditionally created behind `OutOfProcessCommandExecutor.ResolvePwshPath()`, and `PSModulePath` is captured before mutation and restored on dispose. The pre-warm pattern (one discarded `tools/list` followed by the captured one) addresses MAML cache stabilization across runspaces — that detail belongs in the fixture, and it is.
- **Workspace-root resolution by walking up to `PoshMcp.sln`** is duplicated across `HelpParityFixtureSession` and `ToolDescriptionRegressionTests`. Not a blocker, but a future `TestPathHelper` extraction would remove the drift risk.

### Test Strategy
- **FR-520 / FR-521 coverage is solid.** Three dimensions covered: (1) fixture-command-count parity, (2) tool-level description byte-equality across modes for all 6 fixture commands, (3) per-parameter description byte-equality across modes via `Intersect` on tool names. The union-of-keys comparison in `ParameterDescriptions_AreByteIdentical_*` correctly surfaces a missing parameter on one side instead of silently skipping.
- **SC-208 / FR-511 coverage is genuine.** `Resolver_IsInsensitiveToCommandIdentity_GivenSameParameterInputs` is the right invariant — it proves the resolver does not key off command identity, which is the *necessary condition* for "applied per parameter, not per parameter set". `ToolDescription_IsInvariantAcrossParameterSetSyntaxes` adds the FR-501 mirror at the tool level. Determinism via 5x repeat invocation is overkill for a pure function but cheap and self-documenting.
- **FR-550 backward-compatibility regression**: baseline files exist (`baseline/inprocess-tools-list.json` 386 KB, `oop-tools-list.json` 287 KB) and the test asserts the equal-or-superset rule against the committed snapshots. The origin check via live `Get-Help .Synopsis` resolution is the right way to honor the FR-550 carve-out (syntax-fallback / bare-name descriptions are NOT covered).

### Skip Strategy — `[Theory(Skip="Tracking issue #242 …")]` is the right call
- The 10 `ParameterDescription_IsNonEmpty_WhenHelpTextAvailable_*` variants are correctly skipped pinned to #242. They are the **regression gate** — un-skipping them is part of the issue #242 acceptance criteria. The skip messages explicitly point reviewers at the tracking issue.
- Failing them instead would block this PR and every subsequent unrelated PR's CI for a fix that is already triaged, scoped, and assigned its own work item. That's the wrong load-bearing wall — the test addition (measurement) and the wiring fix (remediation) belong in separate PRs with a clear handoff. This PR is the measurement; #242 is the remediation. Standard "tests-then-fix" sequencing with the gate established up-front.
- The two skipped tests appearing in the test report (`Skipped: 2`) plus the explicit skip reason gives the next reviewer (or future-me) a clear breadcrumb without a CI red wall.

### Issue #242 Scope
- **Filing #242 as a separate issue, not blocking #229, is the right architectural choice.** #229 is "add the test trio for spec 010 wave 5" — it's complete the moment the trio exists, the fixture works, and the gate is in place. The wiring gap is a *finding*, not a blocker for the test work. Conflating them would either (a) bloat #229 to also fix the resolver→schema seam (different file, different agent's domain, different review) or (b) leave the gap undocumented and silent.
- Issue #242 is well-scoped: it identifies the exact seam (resolver returns correct strings ↔ MCP `inputSchema.properties.<name>.description` JSON), names the suspected location (SDK schema generator consuming generated assembly's `[Description]` attribute), points at the existing fixture for repro, and lists the 10 specific test methods that un-skip green when the fix lands. Acceptance criteria are crisp.

### One Non-Blocking Nit
- `ToolDescriptionRegressionTests.IsEqualOrSuperset` accepts three cases: exact equality, baseline-prefix-then-paragraph-separator, **and** `current.Contains(baseline)`. The third case is broader than spec FR-550 reads ("synopsis followed by additional text from FR-500 step 2"). `Contains` would accept baseline appearing *anywhere* in the current string, including after additional prefix text. The inline comment calls this "lenient acceptance" and notes it covers post-change synopsis-appended cases — fine pragmatically, but it is a strictness slip vs the spec wording. Suggest either (a) tightening to `EndsWith(ParagraphSeparator + baseline)` to also cover the appended case explicitly, or (b) dropping the `Contains` fallback once a real run confirms it's not needed. Not blocking the PR.

### Summary
Test architecture is correctly layered (one fixture, three consumer test classes, unit vs integration split is right), FR coverage is genuine and not box-ticking, the skip strategy with #242 tracking is the textbook tests-before-fix sequencing, and the baseline-regression mechanism honors the FR-550 carve-out. Approving. The `Contains` fallback in `IsEqualOrSuperset` is the only thing worth tightening, and that is a follow-up nit, not a gate.

— Farnsworth

usepowershell:
## Cubert — Fact-Check

**Verdict: ✅ APPROVE**

### Test Claim Verification

Spot-checked test methods against their declared assertions — names match behavior:

- `ToolDescriptionParityTests.FixtureCommands_AppearInBothModes_WithSameToolVariantCount` — iterates `HelpParityFixtureSession.FixtureCommands`, asserts each has variants in both `_inProcessSession` and `_outOfProcessSession`, and asserts equal variant counts. Does what the name says.
- `ToolDescriptionParityTests.ParameterDescriptions_AreByteIdentical_AcrossRuntimeModes` — matches variants by tool name, computes the **union** of parameter keys (so a missing parameter on one side surfaces as a parity failure rather than silently passing), asserts ordinal string equality on `inputSchema.properties.<name>.description`. Honest.
- `ParameterSetConsistencyTests.HelpParameter_ResolvesIdentically_AcrossParameterSets` — builds a `ParameterDescriptionRequest` with `helpParameterDescription` set, then runs `AssertSameDescriptionAcrossInvocations` with `expectedSource: ParameterDescriptionSource.HelpParameter`. Tests FR-510 step 1 + FR-511 invariant as claimed.
- `ToolDescriptionRegressionTests.AssertEqualOrSuperset` — applies the FR-550 origin check via `OriginatedFromSynopsis` (only asserts on synopsis-sourced descriptions, conservatively skipping non-fixture commands), then evaluates `IsEqualOrSuperset`. Logs `assertedCount` so reviewers can see how many tools the rule actually fired against.

### Skip Strategy & Issue #242 Accuracy

- Skip messages on the 10 `ParameterDescription_IsNonEmpty_*` variants reference issue #242 verbatim and accurately describe the gap: *"resolver returns the right strings but they don't reach inputSchema.properties.<name>.description"*. Reviewers landing on a skipped test can navigate straight to the tracking issue.
- Issue #242 is factually correct against the code:
  - Names the seam precisely — resolver (`HelpAwareToolMetadataSource.ResolveParameterDescription`) returns correct strings per `ParameterSetConsistencyTests` (8 passing facts, verified), but they never reach the generated `inputSchema.properties.<name>.description` JSON.
  - Points at `HelpParityFixture` (6 commands) as the reproducer — matches the fixture corpus actually used.
  - Lists the 10 specific `[Theory(Skip)]` methods as the regression gate — matches the test source.
  - Suspected root cause (`[Description]` attribute attachment in `PowerShellAssemblyGenerator` / `OutOfProcessToolAssemblyGenerator` not being read by the SDK schema generator, OR not being attached for some parameter shapes) is plausibly scoped to PR #241's wiring path.
  - Acceptance criteria are crisp and testable.

### Numbers Verified

I ran the tests in the worktree:

- **23 passed + 2 skipped on new tests:** ✅
  ```
  Passed!  - Failed: 0, Passed: 23, Skipped: 2, Total: 25, Duration: 1m 36s
  ```
  (filter: `FullyQualifiedName~ToolDescriptionParity|FullyQualifiedName~ToolDescriptionRegression|FullyQualifiedName~ParameterSetConsistency`)

  Note: the 2 visible skipped are `[Theory(Skip)]` methods (counted once per theory, not per `[InlineData]` row), which is xUnit's normal reporting. The PR body's "23 passed + 2 skipped" matches what `dotnet test` reports.

- **501/501 unit suite green:** ✅
  ```
  Passed!  - Failed: 0, Passed: 501, Skipped: 0, Total: 501, Duration: 1m 11s
  ```
  (filter: `FullyQualifiedName~Unit`)

### Baseline File Usage

- `specs/010-tool-self-documentation/baseline/inprocess-tools-list.json` — exists.
- `specs/010-tool-self-documentation/baseline/oop-tools-list.json` — exists.
- `ToolDescriptionRegressionTests.LoadBaseline` resolves the workspace root by walking up to `PoshMcp.sln`, then reads from `specs/010-tool-self-documentation/baseline/{filename}.json` and unwraps `result.tools` as a `JArray`. The path matches; the parsing matches the captured snapshot shape (MCP `tools/list` envelope).

### The Gap Itself — Re-verified

The gap I raised in PR #241 is unchanged: every parameter on every tool emits an empty `inputSchema.properties.<name>.description` in both runtime paths. The strongest evidence in this PR is the test split itself — `ParameterDescriptions_AreByteIdentical_AcrossRuntimeModes` passes (parity holds because both modes are empty), while the 10 `ParameterDescription_IsNonEmpty_*` tests are pinned skipped on #242. The two outcomes are only consistent if the descriptions are uniformly empty across modes, which is exactly what the PR #241 verification reported.

### Summary

PR body numbers verify exactly. Test names match behavior. Skip strategy correctly points reviewers at #242. Issue #242 is well-scoped, factually correct against the resolver/codepath, and gives the next agent clear acceptance criteria. Baseline files exist and are loaded correctly. Approving.

Concur with Farnsworth's non-blocking nit on `IsEqualOrSuperset.Contains(baseline)` being broader than the FR-550 spec wording — worth tightening to `EndsWith(ParagraphSeparator + baseline)` in a follow-up, but not a gate for this PR.

— Cubert

PR #248, opened 2026-05-13
fix(metadata): wire parameter descriptions through to MCP inputSchema (#242)

### LINKED ISSUE #242: FR-510 wiring gap: inputSchema parameter descriptions are empty in both InProcess and OOP modes

usepowershell:
## Summary

Fixes #242 — FR-510 wiring gap where MCP `inputSchema.properties.<name>.description` was empty in both in-process and out-of-process modes.

## Root Cause

`PowerShellHelpResolver.ResolveParameterArray` was calling `.BaseObject` on the `Get-Help` `.parameters` PSObject. The PSObject is a wrapper around a `PSCustomObject` marker type with **no public members** — the synthesized `.parameter[]` collection lives on the **wrapper**, not on `BaseObject`. Unwrapping first dereferenced to the empty marker and silently returned no parameters, so every MCP tool's `inputSchema` shipped with empty parameter descriptions.

The resolver had been returning the right strings for help text it could see; the array enumeration step was just dropping the input on the floor before any string lookups happened.

## Fix

```csharp
// Before:
var unwrapped = value is PSObject pso ? pso.BaseObject ?? pso : value;
var holder = unwrapped as PSObject ?? new PSObject(unwrapped);

// After: read on the wrapper first, fall back to BaseObject only if absent.
var holder = value as PSObject ?? new PSObject(value);
var parameterMember = holder.Properties["parameter"]?.Value;
if (parameterMember == null && value is PSObject pso && pso.BaseObject is { } baseObj && !ReferenceEquals(baseObj, pso))
{
    var baseHolder = baseObj as PSObject ?? new PSObject(baseObj);
    parameterMember = baseHolder.Properties["parameter"]?.Value;
}
```

## Secondary Change — `McpToolFactoryV2` Default Source

Three `McpToolFactoryV2` constructors null-coalesced `IToolMetadataSource` to `DefaultToolMetadataSource` (a no-op that returns nothing). In production, `HttpServerHost`/`StdioServerHost` register `HelpAwareToolMetadataSource` via DI, so this default never fired in real runtime. But CLI bootstrappers and any test path that constructs `McpToolFactoryV2` without supplying a metadata source were silently running with help-blind metadata.

Switched all three defaults to `HelpAwareToolMetadataSource` so the contract matches production wiring. **No DI registration changed** — this only affects callers that pass `null` and want sensible defaults, which is what they almost certainly intended.

## Verification

- **`ParameterDescription_IsNonEmpty_*` regression gate (the 10 tests Fry left skipped on #242):** **10/10 passing** — 5 in-process + 5 out-of-process.
- **Broader description surface** (`ToolDescription`, `ParameterSet`, `DoctorDescription`, `DescriptionSource`): **70/70 passing.**
- **Full unit suite:** **532/532 passing**, 0 skipped.

## Files

- `PoshMcp.Server/PowerShell/PowerShellHelpResolver.cs` — root-cause fix
- `PoshMcp.Server/McpToolFactoryV2.cs` — default-source swap (3 ctors)
- `PoshMcp.Tests/Integration/ToolDescriptionParityTests.cs` — un-skipped 10 `[Theory(Skip="…#242…")]` rows; one became `[PwshAvailableTheory]` for the OOP variant

Closes #242.

### DIFF
--- a/.squad/agents/bender/history.md
+++ b/.squad/agents/bender/history.md
@@ -705,3 +705,23 @@ should update to their own path (removed the repo-internal `examples/appsettings
 **#242 observation (FR-510 parameter descriptions not reaching MCP `inputSchema`):** Looked but did not deeply audit. The PR will note this is a separate concern. Hypothesis: `BuildParameterDescriptionMap` does record into the description map, but the description map may not be flowing into the JSON Schema property definition emitted to MCP. Worth its own investigation — different code path from doctor output (which now reads the tracker, not the schema).
 
 **Test coverage (12 unit tests):** all 4 tool sources + all 4 parameter sources via real `HelpAwareToolMetadataSource` resolution + tracker first-wins semantics + JSON round-trip + vocabulary mapping (both enums) + `BuildToolDescriptionEntries` empty/populated paths. 521 unit tests now passing.
+
+## Learnings — 2026-05-13 (Issue #242)
+
+**PowerShell SDK: PSObject wrapper vs BaseObject for Get-Help.parameters**
+
+Get-Help returns .parameters as a PSObject whose BaseObject is a marker PSCustomObject with no public members. The synthesized .parameter[] collection only exists on the **wrapper**, NOT on BaseObject. Calling .BaseObject first dereferences to the marker and silently drops the array.
+
+Rule: when working with PowerShell adapted/synthesized members (especially Get-Help output, format data, custom property sets), access `Properties["name"]` on the PSObject wrapper directly. Only fall back to BaseObject when the wrapper does not expose the member you need. Never reflexively unwrap.
+
+This bug shipped silently because the resolver returned the right strings for parameters it could find, but the array itself was empty — so callers got "no parameters in help" rather than an exception.
+
+## Learnings — 2026-05-13 (Issue #242)
+
+**PowerShell SDK: PSObject wrapper vs BaseObject for Get-Help.parameters**
+
+Get-Help returns .parameters as a PSObject whose BaseObject is a marker PSCustomObject with no public members. The synthesized .parameter[] collection only exists on the **wrapper**, NOT on BaseObject. Calling .BaseObject first dereferences to the marker and silently drops the array.
+
+Rule: when working with PowerShell adapted/synthesized members (especially Get-Help output, format data, custom property sets), access `Properties["name"]` on the PSObject wrapper directly. Only fall back to BaseObject when the wrapper does not expose the member you need. Never reflexively unwrap.
+
+This bug shipped silently because the resolver returned the right strings for parameters it could find, but the array itself was empty — so callers got "no parameters in help" rather than an exception.
--- a/PoshMcp.Server/McpToolFactoryV2.cs
+++ b/PoshMcp.Server/McpToolFactoryV2.cs
@@ -139,7 +139,7 @@ public class McpToolFactoryV2
     public McpToolFactoryV2(IToolMetadataSource? metadataSource, IToolDescriptionSourceTracker? descriptionSourceTracker)
     {
         _assemblyGenerator = new PowerShellAssemblyGenerator(new SingletonPowerShellRunspace());
-        _toolMetadataSource = metadataSource ?? new DefaultToolMetadataSource();
+        _toolMetadataSource = metadataSource ?? new HelpAwareToolMetadataSource();
         _descriptionSourceTracker = descriptionSourceTracker;
     }
 
@@ -172,7 +172,7 @@ public class McpToolFactoryV2
     public McpToolFactoryV2(IPowerShellRunspace runspace, IToolMetadataSource? metadataSource, IToolDescriptionSourceTracker? descriptionSourceTracker)
     {
         _assemblyGenerator = new PowerShellAssemblyGenerator(runspace);
-        _toolMetadataSource = metadataSource ?? new DefaultToolMetadataSource();
+        _toolMetadataSource = metadataSource ?? new HelpAwareToolMetadataSource();
         _descriptionSourceTracker = descriptionSourceTracker;
     }
 
@@ -202,7 +202,7 @@ public class McpToolFactoryV2
     {
         _commandExecutor = commandExecutor ?? throw new ArgumentNullException(nameof(commandExecutor));
         _outOfProcessAssemblyGenerator = new OutOfProcessToolAssemblyGenerator(commandExecutor);
-        _toolMetadataSource = metadataSource ?? new DefaultToolMetadataSource();
+        _toolMetadataSource = metadataSource ?? new HelpAwareToolMetadataSource();
         _descriptionSourceTracker = descriptionSourceTracker;
     }
 
--- a/PoshMcp.Server/PowerShell/PowerShellHelpResolver.cs
+++ b/PoshMcp.Server/PowerShell/PowerShellHelpResolver.cs
@@ -151,10 +151,20 @@ public sealed class PowerShellHelpResolver
 
     private static IEnumerable<object?>? ResolveParameterArray(object value)
     {
-        var unwrapped = value is PSObject pso ? pso.BaseObject ?? pso : value;
-        var holder = unwrapped as PSObject ?? new PSObject(unwrapped);
-
+        // Get-Help returns .parameters as a PSObject wrapping a PSCustomObject. The
+        // synthesized .parameter[] member lives on the PSObject wrapper — calling
+        // .BaseObject would dereference to the PSCustomObject marker type, which has
+        // no public members, dropping the parameter array on the floor (issue #242).
+        // Access Properties directly on the wrapper, only falling back to BaseObject
+        // unwrapping if the wrapper does not expose .parameter.
+        var holder = value as PSObject ?? new PSObject(value);
         var parameterMember = holder.Properties["parameter"]?.Value;
+        if (parameterMember == null && value is PSObject pso && pso.BaseObject is { } baseObj && !ReferenceEquals(baseObj, pso))
+        {
+            var baseHolder = baseObj as PSObject ?? new PSObject(baseObj);
+            parameterMember = baseHolder.Properties["parameter"]?.Value;
+        }
+
         if (parameterMember == null)
         {
             // Some shapes pass the parameter array directly.
--- a/PoshMcp.Tests/Integration/ToolDescriptionParityTests.cs
+++ b/PoshMcp.Tests/Integration/ToolDescriptionParityTests.cs
@@ -210,7 +210,7 @@ public sealed class ToolDescriptionParityTests : PowerShellTestBase, IAsyncLifet
     /// <item>Get-FixtureValidateSetArray.Directions → FR-510 step 3 (array).</item>
     /// </list>
     /// </remarks>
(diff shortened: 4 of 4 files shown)


usepowershell:

**Verdict: ✅ APPROVE**

### Root Cause
Confirmed accurate. Verified `holder.Properties["parameter"]` reads against the PSObject wrapper directly. The pre-fix code unwrapped via `.BaseObject`, which for `Get-Help`'s `.parameters` dereferences to a marker `PSCustomObject` with no public members — the synthesized `.parameter[]` collection lives on the wrapper that adapted view emitted by the help cmdlet, not on the base. Bender's diagnosis is correct and the fix targets the exact line that was dropping the array. The fallback path (try wrapper → if null, try BaseObject as PSObject) is correctly ordered: wrapper first, BaseObject only on miss, and a `ReferenceEquals` guard avoids the infinite-loop case where `BaseObject` returns the same instance.

### Default-Source Swap Risk
Audited all `new McpToolFactoryV2(...)` call sites and the constructor chain. Every parameterless and single-arg overload chains through one of the three constructors that were edited, so the swap propagates everywhere — `PowerShellTestBase`, `WinPsCompatProxyTests`, `ConfigurationReloadTests`, `EndToEndDynamicReloadToolsTests`, `ModuleDiscoveryStartupOrderingTests`, `HttpToolFactoryParityTests`, and `McpToolSetupService` all benefit. Production `HttpServerHost`/`StdioServerHost` DI registration of `HelpAwareToolMetadataSource` is unaffected (those branches don't pass null). The 532/532 unit pass rate confirms no test was depending on the no-op behavior. Low risk.

### DefaultToolMetadataSource Future
**Do not delete in this PR — but it's now dead code in `McpToolFactoryV2`.** Two live references remain in `PowerShellSchemaGenerator.GenerateParameterSchema`/`CreateParameterSchema` (lines 33, 130). Per Cubert's spec-010 audit, `PowerShellSchemaGenerator` is on the documentation-emission cold path and is NOT invoked by `McpToolFactoryV2` for the hot inputSchema path (the MCP SDK reflects on the dynamically-generated assembly). Consistency follow-up: in a separate small PR, swap those two defaults to `HelpAwareToolMetadataSource` as well so `DefaultToolMetadataSource` becomes a single explicit opt-out only. File a tracker; not a blocker for #248.

### In-Proc + OOP Symmetry
Confirmed via test split (5 in-process + 5 out-of-process via `[PwshAvailableTheory]`). Both paths route parameter resolution through `IToolMetadataSource` → `HelpAwareToolMetadataSource` → `PowerShellHelpResolver.ResolveParameterArray`, so the resolver fix benefits both runtimes from a single change. The OOP path was already harvesting Get-Help server-side and shipping resolved fields over ndjson; the resolver fix repairs the in-proc Get-Help consumption that was the actual broken seam.

### Coverage
The 10-row regression gate covers FR-510 steps 1 (`.PARAMETER` block via `Get-FixtureFullHelp.Message` + `.Count`), step 2 (`HelpMessage` attr via `Get-FixtureHelpMessageOnly.UserId`), and step 3 in both scalar and array shapes (`Get-FixtureValidateSetScalar.Color`, `Get-FixtureValidateSetArray.Directions`). Step 4 (typed fallback) is intentionally NOT in this gate — typed fallback is structurally always non-empty and would have masked the original bug. The gate name `_WhenHelpTextAvailable_` correctly scopes the assertion. Coverage is sufficient; the 70/70 broader description surface gives independent corroboration that step 4 still works.

### Summary
Tight, surgical fix with the right comment explaining the PSObject-wrapper invariant for future readers. The default-source swap is a sensible alignment of CLI/test paths with production DI. One non-blocking architectural follow-up: chase the two remaining `DefaultToolMetadataSource` defaults in `PowerShellSchemaGenerator` for consistency once #248 ships. Approving — ready for merge after CI checks complete and Cubert's reviewer pass.

— Farnsworth (Lead Architect), 2026-05-13

usepowershell:
## Cubert — Fact-Check

**Verdict: ✅ APPROVE**

This is the bug I surfaced in PR #241/#243/#244/#245. I held the highest standard. The fix is correct, the regression gate I left as a tracker is now green, and the numbers Bender reported are reproducible byte-for-byte.

### Bug Closure Verified
Re-ran the FR-510 regression gate in the worktree (`poshmcp-242`, branch `fix/issue-242-parameter-descriptions`). All 10 `ParameterDescription_IsNonEmpty_*` rows pass — both in-process and out-of-process. The empty `inputSchema.properties.<name>.description` failure mode is gone.

Root-cause fix is the right shape: `PowerShellHelpResolver.ResolveParameterArray` now reads `Properties["parameter"]` on the wrapper PSObject first, falling back to `BaseObject` only if absent. Comment in the code explains why (PSCustomObject marker has no public members) so this won't be re-introduced. The secondary `McpToolFactoryV2` ctor-default swap (`DefaultToolMetadataSource` → `HelpAwareToolMetadataSource`) closes a parallel CLI/test-bootstrap gap; production DI was already wiring the help-aware source so no behavioural regression for hosted runtimes.

### Numbers Verified
- 10/10 gate: ✅ (5 in-proc + 5 OOP, total 1m 07s)
- 70/70 broader (`ToolDescription|ParameterSet|DoctorDescription|DescriptionSource`): ✅ (2m 41s, 0 skipped)
- 532/532 unit: ✅ (1m 16s, 0 skipped)

### Skip Removal
`grep -r "Skip.*242" PoshMcp.Tests/` in the worktree returns **zero hits**. Both `[Theory(Skip = "Tracking issue #242 …")]` markers from PR #243 are gone — replaced by `[Theory]` (in-proc) and `[PwshAvailableTheory]` (OOP). The `[PwshAvailableTheory]` choice is correct: it preserves the pre-existing pwsh-availability gate without smuggling in a new `Skip` reason.

### End-to-End Trace
Picked `Get-FixtureFullHelp.Message`. Trace: Get-Help `.parameters.parameter[]` (now correctly read on the PSObject wrapper) → `HelpAwareToolMetadataSource` resolves the description → `BuildParameterDescriptionMap` → MCP `tools/list` `inputSchema.properties.Message.description`.

The parity test asserts on exactly the MCP-visible field, not the resolver in isolation:
```csharp
var props = tool["inputSchema"]?["properties"] as JObject;
…
var desc = prop.Value?["description"]?.ToString();
Assert.False(string.IsNullOrWhiteSpace(desc), …);
```
That's the field MCP clients consume. FR-510 contract is enforced end-to-end through the JObject of a real `tools/list` response.

### Closes #242 Keyword
PR body ends with `Closes #242.` — GitHub will auto-close the issue on merge. ✅

### Farnsworth's PowerShellSchemaGenerator Finding
Confirmed accurate. `PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs`:
- Line 33: `var source = metadataSource ?? new DefaultToolMetadataSource();` (in `GenerateParameterSchema`)
- Line 130: `var source = metadataSource ?? new DefaultToolMetadataSource();` (in `CreateParameterSchema` overload)

Both still default to the no-op source. Same shape as the bug Bender just fixed in `McpToolFactoryV2`. Non-blocking for #242 because nothing in the live MCP path calls these — `GenerateParameterSchema` is documented as "for documentation purposes" and the schema actually shipped over MCP comes from `McpToolFactoryV2` (now correctly defaulted). But it's a latent footgun and deserves its own tracker so the next caller of `PowerShellSchemaGenerator` doesn't re-introduce empty descriptions in a doc/diagnostic surface. **+1 to opening a follow-up issue.**

### Minor Observation (non-blocking)
`.squad/agents/bender/history.md` shows the same `## Learnings — 2026-05-13 (Issue #242)` block appended twice (lines 709-714 and 723-728). Cosmetic only — Scribe history is append-only and the duplication doesn't affect anything functional. Bender may want to dedupe in a follow-up commit, or leave it; either is fine.

### Summary
The bug I surfaced is closed. Regression gate I left for #242 is now green and protects against re-introduction. The two-line root cause + the secondary default-source swap are surgical and well-explained. The PR body's `Closes #242` keyword wires the issue closure correctly. Approving with a recommendation to open a tracker for the `PowerShellSchemaGenerator` defaults that Farnsworth flagged. Good fix, Bender. Ship it.

PR #250, opened 2026-05-13
fix(schema): wire HelpAwareToolMetadataSource default in PowerShellSchemaGenerator (#249)

### LINKED ISSUE #249: PowerShellSchemaGenerator still defaults to DefaultToolMetadataSource (FR-510 cold-path follow-up to #242)
## Summary

`PowerShellSchemaGenerator.cs` still defaults `metadataSource` to `DefaultToolMetadataSource` in two places, which returns the typed-fallback string for every parameter and bypasses the spec 010 FR-510 precedence chain (Get-Help → `ParameterAttribute.HelpMessage` → `ValidateSet` phrasing → typed fallback).

This is a **non-blocking follow-up** to #242 / PR #248. The live MCP wire path was fixed in #248 by switching `McpToolFactoryV2`'s default to `HelpAwareToolMetadataSource`. `PowerShellSchemaGenerator` is the cold doc-emission path — not on the live wire — so symptoms are limited to documentation/schema export tooling, not interactive tool calls.

## Locations

- `PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs:33` — `GenerateParameterSchema` default
- `PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs:130` — `CreateParameterSchema` inner default

Both flagged independently by Farnsworth and Cubert during PR #248 review.

## Proposed fix

Swap the `?? new DefaultToolMetadataSource()` defaults to `?? new HelpAwareToolMetadataSource(...)`, mirroring what #248 did for `McpToolFactoryV2`. Verify any callers that currently rely on the no-op fallback either pass an explicit source or accept the upgraded behavior.

## Acceptance

- [ ] Both defaults in `PowerShellSchemaGenerator.cs` use `HelpAwareToolMetadataSource` (or are otherwise wired to honor FR-510)
- [ ] No regression in unit/functional/integration suites (532+ tests still green)
- [ ] If any caller depended on the typed-fallback-only behavior, that call site is updated explicitly

## Context

- Fix PR for the live wire path: #248 (closed #242)
- Spec: `specs/010-tool-self-documentation/`
- Related FRs: FR-500, FR-510

usepowershell:
## Root cause

Both default fallbacks in `PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs` constructed `new DefaultToolMetadataSource()` — the no-op resolver that always returns the typed-fallback string and bypasses the FR-510 precedence chain. Any caller that omitted the optional `metadataSource` argument silently lost HelpMessage / ValidateSet / Get-Help precedence.

This is the cold-path twin of the live-wire fix in PR #248 (issue #242). Same default mistake, different file.

## Changes

`PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs`:

- Line ~33 (`GenerateParameterSchema`): `metadataSource ?? new DefaultToolMetadataSource()` → `metadataSource ?? new HelpAwareToolMetadataSource()`
- Line ~131 (`CreateParameterSchema` overload): same swap
- XML doc updated to describe the new default and the natural degradation when Get-Help text isn't pre-resolved at this layer

`HelpAwareToolMetadataSource` is a parameterless pure resolver — no DI plumbing required. Without pre-resolved Get-Help text the chain degrades naturally to HelpMessage → ValidateSet → typed fallback, which is exactly what the cold path can supply today.

## Caller impact

`grep_search` for `PowerShellSchemaGenerator` / `GenerateParameterSchema` / `CreateParameterSchema` returns matches only inside the file itself. There are **no production or test callers** of this class today, so this is a latent-bug fix (any future doc-emission caller would have hit the same DefaultToolMetadataSource trap that #242 just spent a round diagnosing on the live MCP wire).

## Validation

- `dotnet build PoshMcp.Server\PoshMcp.csproj` — 0 errors, 19 pre-existing warnings (NU1510 + CS8602/CS8604 in untouched files)
- `ParameterDescription_IsNonEmpty` regression gate — **10/10 passed** (5 in-process + 5 OOP)
- Unit suite (`FullyQualifiedName~PoshMcp.Tests.Unit`) — **532/532 passed**

Closes #249

### DIFF
--- a/PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs
+++ b/PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs
@@ -13,8 +13,9 @@ namespace PoshMcp.Server.PowerShell;
 /// <see cref="IToolMetadataSource"/> so they apply the same FR-510 precedence chain
 /// (Get-Help parameter → ParameterAttribute.HelpMessage → ValidateSet phrasing → typed
 /// fallback) used by the in-process and out-of-process tool factory paths. Callers that
-/// don't supply a source receive the type-fallback string for every parameter (preserves
-/// pre-spec-010 output).
+/// don't supply a source get a default <see cref="HelpAwareToolMetadataSource"/> so the
+/// precedence chain still applies on cold doc-emission paths; without pre-resolved
+/// Get-Help text the chain degrades naturally to HelpMessage / ValidateSet / typed fallback.
 /// </remarks>
 public static class PowerShellSchemaGenerator
 {
@@ -23,14 +24,17 @@ public static class PowerShellSchemaGenerator
     /// </summary>
     /// <param name="commandInfo">The PowerShell command information.</param>
     /// <param name="metadataSource">Optional metadata source used to resolve per-parameter
-    /// descriptions per spec 010 FR-510. Defaults to <see cref="DefaultToolMetadataSource"/>
-    /// when omitted, which preserves pre-spec-010 output.</param>
+    /// descriptions per spec 010 FR-510. Defaults to <see cref="HelpAwareToolMetadataSource"/>
+    /// when omitted so the FR-510 precedence chain (Get-Help → HelpMessage → ValidateSet →
+    /// typed fallback) is applied even on cold doc-emission paths. Get-Help inputs are not
+    /// available at this layer, so the chain naturally degrades to HelpMessage / ValidateSet /
+    /// typed fallback unless the caller supplies a pre-resolved source.</param>
     /// <returns>A schema object representing the command parameters.</returns>
     public static object GenerateParameterSchema(CommandInfo commandInfo, IToolMetadataSource? metadataSource = null)
     {
         var properties = new Dictionary<string, object>();
         var required = new List<string>();
-        var source = metadataSource ?? new DefaultToolMetadataSource();
+        var source = metadataSource ?? new HelpAwareToolMetadataSource();
 
         foreach (var parameterKvp in commandInfo.Parameters)
         {
@@ -127,7 +131,7 @@ public static class PowerShellSchemaGenerator
         // Resolve description through the spec 010 precedence chain. Pull the same inputs
         // the in-process call site collects (Get-Help text isn't available here, so the
         // chain naturally degrades to HelpMessage / ValidateSet / typed fallback).
-        var source = metadataSource ?? new DefaultToolMetadataSource();
+        var source = metadataSource ?? new HelpAwareToolMetadataSource();
         var helpMessage = parameterMetadata.Attributes.OfType<ParameterAttribute>()
             .Select(a => a.HelpMessage)
             .FirstOrDefault(m => !string.IsNullOrWhiteSpace(m));


usepowershell:
## ✅ APPROVE

Clean cold-path twin of #248. Both default fallbacks in `PowerShellSchemaGenerator.cs` (`GenerateParameterSchema` and the four-arg `CreateParameterSchema` overload) now construct `HelpAwareToolMetadataSource` instead of `DefaultToolMetadataSource`, restoring the FR-510 precedence chain on the cold doc-emission path. `HelpAwareToolMetadataSource` is a `sealed` pure resolver with the implicit parameterless constructor and never invokes PowerShell, so the cold path can't crash even when `Synopsis`/`HelpParameterDescription` aren't pre-resolved — it degrades naturally to `HelpMessage` → `ValidateSet` → typed fallback, which is exactly what this layer can supply. XML docs on both the class remarks and the `metadataSource` parameter have been updated to describe the new default and the natural degradation. The zero-caller claim verifies independently (`grep_search PowerShellSchemaGenerator.(Generate|Create)` returns nothing in `PoshMcp.Server/**/*.cs` outside the file itself), so this is a latent-bug fix that closes the matching trap any future doc-emission caller would have hit. `ParameterDescription_IsNonEmpty` regression gate (10/10) and unit suite (532/532) both green; CI is 2 successful / 1 skipped / 4 pending with zero failures at time of review.

### Non-blocking

- `HelpAwareToolMetadataSource`'s parameterless ctor is now load-bearing for this file. If a future change adds DI to that resolver, the two default-construction sites here will need a different fallback strategy (don't switch back to `DefaultToolMetadataSource`). Worth a comment near the swap or an architectural note in the resolver itself if it grows dependencies.
- The two-arg `CreateParameterSchema(ParameterMetadata)` overload still passes `metadataSource: null`, which now lands on the new default — correct behavior, but the comment on L67-68 could be clarified to call out that it inherits the HelpAware default rather than implying a no-op.

usepowershell:
## ✅ APPROVE

Independent verification of PR #250 — cold-path twin of #248.

### Diff sanity
`gh pr diff 250` shows exactly two `DefaultToolMetadataSource` → `HelpAwareToolMetadataSource` swaps in `PoshMcp.Server/PowerShell/PowerShellSchemaGenerator.cs` (one in `GenerateParameterSchema` at L36, one in the four-arg `CreateParameterSchema` overload at L134) plus XML doc updates on the class remarks and the `metadataSource` parameter. `+10 -6`, single file. Matches the description.

### Zero-caller claim — verified
Independent grep across `PoshMcp.Server`, `PoshMcp.Tests`, `PoshMcp.Benchmarks`, and `TestClient`:
```
Get-ChildItem -Recurse -Include *.cs -Path PoshMcp.Server,PoshMcp.Tests,PoshMcp.Benchmarks,TestClient |
  Select-String "PowerShellSchemaGenerator"
```
returns one path: `PoshMcp.Server\PowerShell\PowerShellSchemaGenerator.cs` (the file itself). No external callers in production or tests. Latent-bug fix as described.

### Gate tests — re-run, observed counts
- `dotnet build PoshMcp.Server\PoshMcp.csproj` — 0 errors, 19 pre-existing warnings (NU1510 + CS8602/CS8604 in untouched files). Matches.
- `dotnet test --filter "FullyQualifiedName~ParameterDescription_IsNonEmpty"` — **10 / 10 passed** (5 InProcessMode + 5 OutOfProcessMode), 1m 30s. Matches.
- `dotnet test --no-build --filter "FullyQualifiedName~PoshMcp.Tests.Unit"` — **532 / 532 passed**, 1m 22s. Matches.

### CI
`gh pr checks 250` at review time: 6 successful, 1 pending (CI/build), 0 failing. CodeQL (actions/csharp/python), Squad CI, and Dependency Submission all green.

### Counter-hypothesis
Could the swap regress anything? `HelpAwareToolMetadataSource` is a sealed pure resolver that never invokes PowerShell — without pre-resolved Get-Help text it degrades to the same precedence chain (HelpMessage → ValidateSet → typed fallback) the cold path could already produce, just with the FR-510 ordering applied. No new failure modes. No external callers means no behavior change for shipped code paths.

### Discrepancy with Bender's report
None. Build, gate, and unit counts reproduce exactly. Zero-caller claim reproduces exactly.

Fine. It checks out.

PR #253, opened 2026-05-14
docs(009): add TESTING.md with per-category commands

usepowershell:
Adds `TESTING.md` at the repo root documenting per-category local commands for the spec 009 test-suite categorization rollout.

**Closes #214.** Covers FR-408. References FR-413, FR-417.

## What's in TESTING.md

- **Category table** — all six categories (`Unit`, `Integration`, `OutOfProcess`, `Http`, `Azure`, `Functional`) with what each covers, typical duration, and whether they spawn `pwsh` / bind ports / require credentials.
- **Per-category command** — exact `dotnet test --filter "Category=<X>"` invocation that reproduces the corresponding CI phase.
- **Per-category caveats** — `Unit` hygiene contract (FR-401/402/403), `Azure` skip-when-no-creds behavior (FR-413), `OutOfProcess` cleanup expectations (FR-412), `Http` dynamic port requirement (FR-411), `Functional` external-resource exclusion rule (FR-416).
- **Default bucket section** — documents the FR-417 contract: untagged tests fall back to a permissive bucket that is **not** `Unit`. Concrete bucket name is deferred to `PoshMcp.Tests/README.md` (set by #212) so this doc stays correct against any choice landed there.
- **CI phase → local command** mapping table, plus a note that the workflow file is the source of truth for actual phase order (#215).
- **Troubleshooting** — common failure modes (Unit tests starting to spawn `pwsh`, `Http` "address already in use", `Azure` running unintentionally, folder-vs-trait confusion).

## Cross-references

This PR is part of the spec 009 milestone and runs in parallel with:

- **#212** — Fry's PR adding the actual `[Trait("Category", ...)]` attributes and the default-bucket implementation. TESTING.md is written against the spec's contract, not against #212's specific bucket name. Cubert will re-verify the doc against the landed implementation.
- **#215** — Amy's PR rewiring the CI workflow to phased category-scoped runs. TESTING.md's "Reproducing a CI phase locally" section maps each phase to a local command; the workflow file remains the source of truth for actual phase order and runner configuration.

## What changed in README.md

The existing `### Testing` section was rewritten to:
- Replace the old folder-based examples (`--filter "FullyQualifiedName~Unit"`) with the trait-based form (`--filter "Category=Unit"`).
- Keep a quick-start (Unit + 3 examples) inline.
- Defer the full reference to `TESTING.md` via a single prominent link.

No duplication of the category table between README and TESTING.md.

## Voice

Neutral, professional documentation throughout — no Squad/agent voice in the published file.

[...]

PR #252, opened 2026-05-14
ci(009): run tests as category-scoped phases

usepowershell:
## Summary

Replaces the single Unit + Functional test job in `.github/workflows/ci.yml` with a sequence of **category-scoped phases** per spec 009, so a CI failure is attributable to exactly one category in the UI.

**Phase order (each is its own step with its own duration / pass-fail summary):**

| Phase          | Filter expression          | TRX log                  |
| -------------- | -------------------------- | ------------------------ |
| Unit           | `Category=Unit`            | `phase-unit.trx`         |
| Integration    | `Category=Integration`     | `phase-integration.trx`  |
| OutOfProcess   | `Category=OutOfProcess`    | `phase-outofprocess.trx` |
| Http           | `Category=Http`            | `phase-http.trx`         |
| Functional     | `Category=Functional`      | `phase-functional.trx`   |
| Azure          | `Category=Azure`           | `phase-azure.trx`        |

`continue-on-error` is left at its default (`false`), so a phase failure halts subsequent phases — matches today's serial-suite behavior.

## Azure skip mechanism

The Azure phase is **skipped (not failed)** when `AZURE_CLIENT_ID` is absent:

- A job-level env `HAS_AZURE_CREDS: ${{ secrets.AZURE_CLIENT_ID != '' }}` projects secret presence into an evaluable expression (GitHub Actions does not allow direct `secrets.X != ''` in step-level `if:` outside environment contexts).
- The Azure step runs only when `env.HAS_AZURE_CREDS == 'true'`.
- A companion "skipped — no credentials" step echoes a message when the secret is absent, so the UI shows an explicit "skipped" status rather than a missing step.

Credentialed Azure CI is deferred per spec 009.

## Dependency note — issue #212 (Fry, in flight)

This workflow uses `--filter "Category=X"` filters. Issue #212 is rolling out the actual `[Trait("Category", ...)]` attributes across the test suite in parallel. Until those land:

- Tests **with** the trait route to the right phase.
- Tests **without** the trait fall into the default bucket and are excluded from category-scoped phases — those phases still run honestly against whatever is currently labeled.
- As traits land, each phase automatically picks up its newly-tagged tests.

This branch does **not** block on #212.

## Acceptance (spec 009 §FR-409, SC-104, SC-106, FR-413)

- ✅ Distinct phase summaries with separate durations (each phase = own step in CI UI)
- ✅ A failed phase produces a CI status whose failure message identifies the category (`Test (OutOfProcess)`)
[...]

PR #257, opened 2026-05-14
ci(009): add flake-rate measurement workflow

### LINKED ISSUE #216: CI: dedicated flake-rate measurement step
# Issue 5 — CI: dedicated flake-rate measurement step

**Title:** CI: dedicated flake-rate step that re-runs the phased suite N times and reports a summary

**Labels:** `squad`
**Milestone:** Spec 009: Test Suite Consistency

## Problem

"0 flakes across 5 runs on `main`" (SC-105, FR-405) is an acceptance bar with no measurement mechanism today. Manual log inspection is unreliable and unrepeatable. Per the resolution of OQ-7, this work needs a dedicated CI step.

## Scope

- Add a CI job (separate workflow or a separately-triggered job in `ci.yml`) that:
  - Runs the full phased suite N times in sequence on a clean runner.
  - Initial N = 5 (configurable as a workflow input or env var).
  - Triggered on a schedule (e.g., nightly on `main`) and on-demand via workflow_dispatch.
- Aggregates the N runs into a single flake-rate summary artifact (JSON or markdown) that lists:
  - Per-test failure count across the N runs (e.g., `TestX: 2/5 runs failed`).
  - Overall flake rate (e.g., `total flaky test instances / total test runs`).
  - The base commit SHA and runner metadata.
- The artifact is uploaded as a build artifact (or posted as a workflow summary) so it's reviewable without re-running the suite.

## Acceptance

- [ ] A CI workflow exists that runs the phased suite N = 5 times in a single job/run.
- [ ] A single artifact is produced per run that reports flake counts per test across the N runs.
- [ ] The workflow can be triggered manually via `workflow_dispatch`.
- [ ] The workflow runs on a schedule on `main` (suggested: nightly).
- [ ] Documentation in `TESTING.md` (or alongside it) points to where to find the latest flake-rate artifact.

## Spec references

Covers **FR-418** (the OQ-7 resolution), supports **FR-405, SC-105**.

Spec: [specs/009-test-suite-consistency/spec.md](../blob/main/specs/009-test-suite-consistency/spec.md)

usepowershell:
## Summary

Adds `.github/workflows/flake-rate.yml` — a dedicated CI workflow that re-runs the phased test suite N times back-to-back on a clean runner and emits a single human-readable flake-rate summary.

Closes #216. Implements spec 009 **FR-418** and provides the measurement substrate for **FR-405** / **SC-105** (zero-flake target across 5 consecutive runs on `main`).

## What it does

- **Triggers**
  - `workflow_dispatch` with a `runs` input (default `5`) so a maintainer can crank N up on demand.
  - `schedule: '0 7 * * *'` — nightly at 07:00 UTC, off-peak versus on-demand CI.
- **Loop:** one job runs N iterations of the phased sequence. Each iteration writes its TRX outputs to `flake-runs/run-${i}/`. `set +e` plus per-phase exit codes captured to `exit-codes.txt` so a flaky phase in iteration 3 does **not** short-circuit iterations 4 and 5.
- **Phasing:** Unit → Integration → OutOfProcess → Http → Functional. Mirrors PR #252 (`ci.yml`) one-for-one. Azure phase is intentionally excluded — scheduled runs have no creds and Azure tests already skip-when-no-creds, so including the phase would only add noise.
- **Aggregator** (PowerShell, matches repo idiom): walks every TRX via `Select-Xml` + `XmlNamespaceManager`, counts non-Passed outcomes, breaks out `Failed` / `NotExecuted` / `Other` separately. Writes `flake-rate-summary.md` with:
  1. Run metadata (UTC timestamp, commit SHA, ref, runner, .NET SDK, totals, aggregate flake rate, run URL).
  2. Per-test failure-count table for any test that did not pass in at least one iteration, sorted by total non-pass count, with iteration list.
  3. Per-iteration phase exit-code matrix so a fully red iteration (env issue) reads differently from a single phase flicking red across iterations (classic flake).
- **Artifacts**
  - `flake-rate-summary` → the markdown summary (the headline artifact).
  - `flake-runs-raw` → full `flake-runs/` tree with per-iteration TRX for drill-down.
  - The summary is also mirrored into `$GITHUB_STEP_SUMMARY` so reviewers can read it on the run page without downloading.

**Aggregate rate definition:** `total non-pass instances / total test invocations across all iterations`. This intentionally counts repeated failures of the same test in different iterations as separate flake instances — one test failing 3/5 weighs more than three different tests each failing 1/5.

## Acceptance check (FR-418)

- [x] CI workflow exists that runs the phased suite N=5 times in a single run.
- [x] A single artifact reports flake counts per test across the N runs.
- [x] Triggerable via `workflow_dispatch` (with configurable `runs` input).
- [x] Schedule on `main` (nightly).
- [x] Summary appended to `$GITHUB_STEP_SUMMARY` for inline visibility.

## Cross-PR coordination

- **PR #252 (`ci.yml`).** This workflow's phasing matches `ci.yml` from PR #252 one-for-one. If #252 lands first, both files reference the same phase definitions implicitly. If #252 has not landed when this PR ships, `flake-rate.yml` still works — it is a separate workflow file with no dependency on `ci.yml` existing. Either order is safe.
- **PR #253 (`TESTING.md`).** TESTING.md is being introduced in PR #253 and is **not** edited from this branch. When #253 is merged, the TESTING.md "Triage" / "Where to find flake data" section should add a pointer like:
  > **Flake-rate measurement.** Flake counts across N consecutive runs of the phased suite are produced by the `Flake Rate Measurement` workflow. Look for the `flake-rate-summary` artifact on a recent run, or read the rendered summary directly on the workflow run page.
  
  Whoever lands #253 (or a follow-up PR) should add that pointer; nothing here is gated on it.

## Why a separate workflow file rather than extending `ci.yml`

- A flake-rate run is fundamentally a different shape (loops + aggregator) than a normal CI build (single pass, fail fast). Forcing them into one workflow either bloats every PR build or hides the loop behind a job conditional that is harder to reason about.
- Separate workflow → separate trigger surface (`workflow_dispatch` + `schedule`) without polluting the on-push/on-PR triggers in `ci.yml`.
- Independent runtime budget — a 5x phased run is multi-multiple of a normal CI run; isolating it keeps PR CI snappy.

## Files touched

- `.github/workflows/flake-rate.yml` (new)
- `.squad/agents/amy/history.md` (Learnings appended)
- `.squad/decisions/inbox/amy-spec009-216-flake-rate.md` (decision drop, will be merged into `decisions.md` by Scribe)

### DIFF
--- a/.github/workflows/flake-rate.yml
+++ b/.github/workflows/flake-rate.yml
@@ -1,0 +1,345 @@
+name: Flake Rate Measurement
+
+# Spec 009 (issue #216 / FR-418, supports FR-405 and SC-105):
+# Re-runs the phased test suite N times back-to-back on a clean runner and
+# aggregates the per-test failure counts into a single human-readable
+# flake-rate summary. Triggered nightly on main and on demand.
+#
+# Phasing matches `.github/workflows/ci.yml` (PR #252 / issue #215). If that
+# workflow has not landed yet, this workflow still runs — it owns its own
+# phased sequence and does not depend on `ci.yml` being present. If #252 is
+# in flight on a different branch, both workflows can co-exist; this one is
+# the source of truth for flake-rate measurement only.
+#
+# Azure phase is intentionally excluded:
+#   - Scheduled / on-demand runs do not have Azure credentials wired
+#     (per spec 009 Non-Goals — credentialed CI for Azure is deferred).
+#   - Azure tests skip-when-no-creds, so including the phase would just
+#     add noise without adding signal.
+
+on:
+  workflow_dispatch:
+    inputs:
+      runs:
+        description: 'Number of suite repetitions (default 5).'
+        required: false
+        default: '5'
+        type: string
+  schedule:
+    # Nightly at 07:00 UTC. Off-peak vs. typical maintainer working hours
+    # so a long measurement run does not contend with on-demand CI.
+    - cron: '0 7 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  measure-flake-rate:
+    runs-on: ubuntu-latest
+
+    # RUN_COUNT defaults to 5 for the nightly schedule; workflow_dispatch
+    # overrides via the `runs` input. The aggregator parses whatever
+    # iteration directories actually exist, so a partial run (e.g. one
+    # iteration crashed early) still produces a usable summary.
+    env:
+      RUN_COUNT: ${{ github.event.inputs.runs || '5' }}
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: 10.x
+
+    - name: Restore dependencies
+      run: dotnet restore
+
+    - name: Build
+      run: dotnet build --configuration Release --no-restore
+
+    # ------------------------------------------------------------------
+    # Repetition loop.
+    #
+    # One bash loop runs RUN_COUNT iterations. Each iteration runs the
+    # same phased sequence as ci.yml (Unit -> Integration -> OutOfProcess
+    # -> Http -> Functional). Per-iteration TRX files land in
+    # `flake-runs/run-${i}/` so the aggregator can attribute every result
+    # to the iteration it came from.
+    #
+    # `set +e` is critical: a phase failing in iteration 3 must NOT skip
+    # iterations 4 and 5 — the whole point of the workflow is to measure
+    # flakes across ALL iterations. We capture per-phase exit codes into
+    # iteration-scoped status files so the aggregator can report which
+    # phases failed in which run, but the loop itself never short-circuits.
+    # ------------------------------------------------------------------
+    - name: Run phased suite N times
+      shell: bash
+      run: |
+        set +e
+        mkdir -p flake-runs
+        echo "Running suite ${RUN_COUNT} times."
+        for i in $(seq 1 "${RUN_COUNT}"); do
+          out="flake-runs/run-${i}"
+          mkdir -p "${out}"
+          echo "::group::Iteration ${i} of ${RUN_COUNT}"
+
+          dotnet test --no-build --configuration Release \
+            --filter "Category=Unit" \
+            --logger "trx;LogFileName=phase-unit.trx" \
+            --results-directory "${out}"
+          echo "unit=$?" >> "${out}/exit-codes.txt"
+
+          dotnet test --no-build --configuration Release \
+            --filter "Category=Integration" \
+            --logger "trx;LogFileName=phase-integration.trx" \
+            --results-directory "${out}"
+          echo "integration=$?" >> "${out}/exit-codes.txt"
+
+          dotnet test --no-build --configuration Release \
+            --filter "Category=OutOfProcess" \
+            --logger "trx;LogFileName=phase-outofprocess.trx" \
+            --results-directory "${out}"
+          echo "outofprocess=$?" >> "${out}/exit-codes.txt"
+
+          dotnet test --no-build --configuration Release \
+            --filter "Category=Http" \
+            --logger "trx;LogFileName=phase-http.trx" \
+            --results-directory "${out}"
+          echo "http=$?" >> "${out}/exit-codes.txt"
+
+          dotnet test --no-build --configuration Release \
+            --filter "Category=Functional" \
+            --logger "trx;LogFileName=phase-functional.trx" \
+            --results-directory "${out}"
+          echo "functional=$?" >> "${out}/exit-codes.txt"
+
+          echo "::endgroup::"
+        done
+        # Surface exit code 0 to the loop step itself — failures are
+        # captured in exit-codes.txt and the TRX results, not via the
+        # shell exit code. The aggregator decides whether to fail the job.
+        exit 0
+
+    # ------------------------------------------------------------------
+    # Aggregator.
+    #
+    # PowerShell, because (a) this repo lives in pwsh land and (b)
+    # `Select-Xml` is a clean way to walk TRX without dragging in a
+    # third-party parser. The summary is markdown rather than JSON
+    # because the primary consumer is a human looking at the workflow
+    # run page (we append it to $GITHUB_STEP_SUMMARY).
+    #
+    # Output: flake-rate-summary.md
+    #   - Run metadata (commit SHA, runner OS, .NET SDK, RUN_COUNT,
+    #     start/end timestamps inferred from TRX).
+    #   - Aggregate flake rate (total flaky test instances / total test
+    #     runs across iterations).
+    #   - Per-test failure-count table for any test that failed at least
+    #     once across the N iterations.
+    #   - Per-iteration phase exit codes (so a fully-failed iteration
+    #     stands out from a partially-flaky one).
+    # ------------------------------------------------------------------
+    - name: Aggregate flake rate
+      if: always()
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = 'Stop'
+        $runRoot = Join-Path $PWD 'flake-runs'
+        $summary = Join-Path $PWD 'flake-rate-summary.md'
+
+        if (-not (Test-Path $runRoot)) {
+          @"
+        # Flake Rate Summary
+
+        No iterations executed. The repetition step did not produce a
+        ``flake-runs/`` directory. See the workflow log for the failure.
+        "@ | Set-Content -Path $summary -Encoding utf8
+          return
+        }
+
+        # Iteration directories are run-1, run-2, ... — sort numerically
+        # so the per-iteration table reads in execution order, not
+        # lexicographic order (run-10 vs run-2).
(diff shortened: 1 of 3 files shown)


usepowershell:
# Cubert — Fact-Check Verdict: PR #257

**PR:** `ci(009): add flake-rate measurement workflow` (Amy) — closes #216
**Branch:** `squad/216-ci-flake-rate-measurement`
**Date:** 2026-05-14
**Verdict:** ✅ **APPROVE** — all claims verified against the actual workflow file.

## Facts verified

| Claim | Result | Evidence |
|---|---|---|
| `workflow_dispatch` with `runs` input, default `5`, type `string` | ✅ | YAML parse: `inputs: {'runs': {'description': 'Number of suite repetitions (default 5).', 'required': False, 'default': '5', 'type': 'string'}}` |
| `schedule: '0 7 * * *'` = nightly 07:00 UTC | ✅ | Cron field parsed as `[{'cron': '0 7 * * *'}]`. Standard 5-field cron `min hr dom mon dow` → 07:00 UTC every day. Matches "nightly UTC". |
| 5 phases mirrored from #252: Unit → Integration → OutOfProcess → Http → Functional, Azure excluded | ✅ | Five `dotnet test` invocations in the loop, in that exact order, with no Azure step. Comment at top of file calls out the Azure exclusion with reason. |
| Trait filter syntax `Category=X` matches PR #252's `ci.yml` | ✅ | Diff of #252 shows `--filter "Category=Unit"`, `Category=Integration`, `Category=OutOfProcess`, `Category=Http`, `Category=Functional`. PR #257 uses the same five strings verbatim. |
| Aggregator parses TRX into per-test failure counts | ✅ | Walks `flake-runs/run-*/*.trx`, uses `Select-Xml` equivalent (`SelectNodes('//t:UnitTestResult', $ns)` with `XmlNamespaceManager` bound to `http://microsoft.com/schemas/VisualStudio/TeamTest/2010`), increments `Failed`/`NotExecuted`/`Other` per test name when `outcome != 'Passed'`. |
| Two artifacts uploaded: `flake-rate-summary` and `flake-runs-raw` | ✅ | Two `actions/upload-artifact@v4` steps with those exact names, both `if: always()`. |
| `$GITHUB_STEP_SUMMARY` mirrored | ✅ | `Get-Content -Path $summary -Raw | Add-Content -Path $env:GITHUB_STEP_SUMMARY -Encoding utf8` after the summary is written. |
| YAML re-validated independently | ✅ | Python `yaml.safe_load` on raw file from `gh api .../contents/...?ref=squad/216-ci-flake-rate-measurement` → `PARSED OK`. Triggers, inputs, and cron all parse to the expected shapes. |
| FR-418: runs phased suite N times (initial N=5) and reports a single summary artifact identifying which tests failed in any of N runs and how often | ✅ | Loop runs `RUN_COUNT` iterations (defaults to 5 via `${{ github.event.inputs.runs || '5' }}`). Aggregator emits `flake-rate-summary.md` with per-test failure-count table — sorted by total non-pass count, lists which iterations the test failed in. Single artifact (`flake-rate-summary`) carries the headline data. |
| Workflow is a new file, not a modification of `ci.yml` | ✅ | `gh pr view` shows `.github/workflows/flake-rate.yml` with `changeType: ADDED`, additions 345 / deletions 0. `ci.yml` is not in the file list. |
| `set +e` + per-phase exit codes captured to `exit-codes.txt` so a flaky phase doesn't short-circuit later iterations | ✅ | First line of bash step is `set +e`; each `dotnet test` is followed by `echo "<phase>=$?" >> "${out}/exit-codes.txt"`; loop ends with explicit `exit 0`. |

## Notes (non-blocking)

- The aggregator's sort key (`Sort-Object { -([int]$_.Value.Failed + ...) }`) uses arithmetic negation for descending order. Works correctly; just unusual style. Not a defect.
- Iteration directories are sorted numerically (`Sort-Object { [int]($_.Name -replace '^run-','') }`), so `run-10` lands after `run-2` — correct and explicitly noted in a comment.
- Phasing comment correctly identifies the dependency story with PR #252 — both files can co-exist; neither blocks the other.

## Decision

**APPROVE.** Workflow YAML parses cleanly, trigger/input/cron definitions match PR claims exactly, phase filter strings align with PR #252's `ci.yml` one-for-one, aggregator implements FR-418 as described, and both artifacts plus the `$GITHUB_STEP_SUMMARY` mirror are present. Ready to merge.

usepowershell:

**Verdict: ✅ APPROVE**

### Workflow Shape
Separate workflow file is the right call. A flake-rate run loops + aggregates;
a normal CI run is single-pass + fail-fast. Forcing both into `ci.yml` would
either bloat every PR build with a conditional or hide the loop behind a job
gate that's hard to reason about. Independent trigger surface
(`workflow_dispatch` + nightly `cron: '0 7 * * *'`) keeps the on-push/on-PR
budget in `ci.yml` clean. `permissions: contents: read` is minimal-correct.

### Trigger / Iteration Semantics
N=5 default with `workflow_dispatch.inputs.runs` override resolved via
`${{ github.event.inputs.runs || '5' }}` — works for both schedule and
dispatch paths. Iteration loop uses `set +e` and captures per-phase exit
codes to `exit-codes.txt` so a failure in iteration 3 doesn't short-circuit
4 and 5. Loop step ends with explicit `exit 0` so the aggregator owns the
gating decision, not the shell. This is the correct shape for measurement.

### Phase Mirroring (FR-407 / PR #252)
Phasing is Unit → Integration → OutOfProcess → Http → Functional, one-for-one
with PR #252's `ci.yml`. No matrix, no parallel matrix legs, no `xunit.parallel`
flips — FR-407's `DisableTestParallelization = true` invariant is preserved.
Azure phase intentionally excluded — credentialed CI is a deferred non-goal
and Azure tests skip-when-no-creds anyway, so including the phase would be
pure noise. Justification documented inline in the workflow header.

### Measurement vs. Gating Divergence (Deliberate, Correct)
Within an iteration, all five phases run regardless of earlier phase failure
(via `set +e`). Production `ci.yml` halts on first phase failure. This is
intentional and correct for FR-418: flake measurement wants maximum signal,
including data on tests that production CI would never reach because an
earlier phase failed. A test that's "always green" only because the suite
halts before reaching it is exactly the kind of flake this workflow is built
to surface. Worth pointing at from TESTING.md when #253 lands; nothing here
is gated on it.

### Aggregator Robustness
PowerShell + `Select-Xml` with an `XmlNamespaceManager` bound to
`http://microsoft.com/schemas/VisualStudio/TeamTest/2010` — correct, TRX
uses a default namespace and dotted-property access silently returns nothing
otherwise. Defensive paths verified:
- Missing `flake-runs/` → emits a stub summary explaining the no-iteration
  case rather than crashing.
- TRX parse failure → `try/catch` logs the file and continues; one corrupt
  TRX from a phase that crashed mid-write does not poison the whole report.
- Missing `exit-codes.txt` lines → per-phase column renders as `-`, so a
  partial iteration is visually distinct from a fully-red one.
- Iteration directory sort uses `[int]($_.Name -replace '^run-','')` → run-10
  appears after run-9, not after run-1.
- `if: always()` on the aggregator and both upload steps → summary and raw
  TRX are produced even when the iteration loop blows up.

### Artifact Contents
`flake-rate-summary.md` carries: UTC timestamp, commit SHA, ref, runner,
.NET SDK, iterations completed vs. requested, total invocations, total
non-pass count, aggregate flake rate, run URL, per-test failure-count
table (only tests that failed at least once, sorted by total non-pass
descending, with iteration list), and per-iteration phase exit-code
matrix. The phase matrix is the right diagnostic primitive — fully red
row → environmental issue; one phase flicking red across rows → classic
flake. Aggregate rate definition (instance count, not unique-test count)
is correct: one test failing 3/5 SHOULD weigh more than three tests each
failing 1/5. `flake-runs-raw` artifact preserves per-iteration TRX for
drill-down. No path collisions across iterations or phases.

### `$GITHUB_STEP_SUMMARY` Sizing
GitHub caps step summary at 1 MB. The per-test table only includes
non-passing tests, so the catastrophic worst case (every test flaked
every iteration) is the only path to overflow — and at that point the
workflow has bigger problems than a truncated summary. Realistic
operating range (a handful of flaky tests across 5 iterations) is well
under the cap.

### Minor Observations (No Action Required)
- No `concurrency:` group set. Two overlapping runs (manual fired during
  a still-running nightly) would run in parallel. Low probability with a
  ~hours-long single-job workflow, but adding `concurrency: { group:
  flake-rate, cancel-in-progress: false }` would make that explicit.
  Not a blocker — flagging for if-you-touch-this-again.
- Idempotency confirmed: rerunning the workflow on the same commit
  rebuilds `flake-rate-summary.md` from scratch each time and uploads
  fresh artifacts. No accumulating state.

### Summary
Workflow design is clean, aggregator is defensively coded, artifacts
match FR-418's "single summary identifying which tests failed and how
often" language, scheduling covers SC-105's "5 consecutive runs on main"
acceptance path, and FR-407's no-parallelism invariant is preserved.
Phase mirroring tracks PR #252 one-for-one. Approving as draft; flip to
ready-for-review and merge.

— Farnsworth

PR #271, opened 2026-05-15
feat(011): OOP wire-format parity for moduleImports (Phase 2 of #268)

usepowershell:
## Summary

Phase 2 of spec 011 closes #268 by extending the out-of-process discover wire format so the C# server can build the doctor `moduleImports` section from data the OOP host already produced — instead of re-running `Get-Module -ListAvailable` in-process. This is the contract that achieves **SC-263-3** (byte-identical `moduleImports` JSON across `InProcess` and `OutOfProcess` runtime modes).

Pairs with #269 (Phase 1 in-process `ModuleDiscovery`) and #270 (Bender's consumer wiring) to close out spec 011 / #267 / #268.

## Wire format additions (additive, backward compatible)

- **`RemoteToolSchema`** gains three nullable string fields per FR-263-9: `SourceModule`, `SourcePattern`, `SourceDetail`. Older OOP hosts that omit these deserialize cleanly with null values per Newtonsoft.Json defaults (**SC-263-4**). Per-command attribution priority: `commandName` > `module` > `pattern`.
- **New top-level `RemoteModuleImportsPayload`** on the `discover` response carries per-module probe results (`Name`/`Found`/`Version`/`Path`) and per-pattern match data (`Pattern`/`Kind`/`Role`/`MatchedCount`).
- Both **`oop-host.ps1`** (single host) and **`oop-host-pool.ps1`** (runspace pool) populate the new fields and emit the payload. The pool wraps the script-block return value as a `PSCustomObject {Schemas, ModuleImports}` and the outer Invoke handler unwraps with a defensive fallback for legacy bare-array shape.

## C# consumer side

- **`ICommandExecutor.LastModuleImports`** default-implementation property returns null. `OutOfProcessCommandExecutor` and `OutOfProcessSubprocessPool` parse the optional payload after the schemas, stash it, and expose it via the property (with thread-safe accessor on the pool, cleared on fingerprint mismatch).
- **`McpToolSetupService.DiscoverToolsAsync`** captures the executor's `LastModuleImports` into a new `OopModuleImportsCapture` `AsyncLocal<RemoteModuleImportsPayload?>` helper before the executor lease disposes. AsyncLocal chosen over signature refactor because CLI doctor invocations are one-shot (no cross-invocation leak risk) and the alternative would have meant changing the public-ish `DiscoverToolsForCliAsync` return shape.
[...]

PR #269, opened 2026-05-15
feat(011): in-process ModuleDiscovery helper (Phase 1 of #268)

usepowershell:
## Summary

**Phase 1 of #268** — adds the in-process module discovery helper. OOP wire-format parity (Phase 2) ships in a follow-up PR per the split allowance in the issue body ("If scope is too large for one PR, land helper + in-process impl first, defer OOP wire-format to a follow-up PR under #268 with checklist.").

### `PoshMcp.Server.PowerShell.ModuleDiscovery`

A static helper that probes each configured PowerShell module via `Get-Module -ListAvailable -Name <name>` and returns one `ModuleProbeResult(Name, Found, Version, Path)` per input entry.

```csharp
public static IReadOnlyList<ModuleProbeResult> ProbeModules(
    IPowerShellRunspace runspace,
    IReadOnlyList<string>? moduleNames,
    ILogger? logger = null);
```

**Contract enforced by tests:**
- ✅ FR-263-10 — exactly one `Get-Module` call per configured module name (never per command)
- ✅ Reuses the supplied `IPowerShellRunspace` — never spawns a new `pwsh` process
- ✅ No side effects (uses `-ListAvailable`, never imports)
- ✅ Probe failures are logged at warning level and yield `Found=false` — never throw

### Tests

10 new unit tests in `PoshMcp.Tests/Unit/ModuleDiscoveryTests.cs`:
- Null / empty / blank-only inputs → empty result
- Non-existent module → `Found=false`, null version & path
- Real built-in module (`Microsoft.PowerShell.Management`) → `Found=true` with valid version + existing `ModuleBase` directory
- Input order preserved across mixed found/missing names
- Whitespace trimmed from names
- Blank entries skipped while valid ones still probed
- Duplicate names → one result per input entry (per spec)
- Null runspace → `ArgumentNullException`

All pass: `Passed: 10, Failed: 0, Duration: 2 s`.

## Why split?

The full spec 011 work touches three independent surfaces:
1. **In-process helper** (this PR) — small, self-contained, immediately useful
2. **OOP wire-format** — additive `SourceModule`/`SourcePattern`/`SourceDetail` on `RemoteToolSchema`, source attribution in both `oop-host.ps1` and `oop-host-pool.ps1`, parallel `moduleImports` payload on the discover RPC response, and backward-compat fallback (older OOP host → `source: "unknown"` + one-time warning)
3. **Parity tests** (SC-263-3) — cross-checks that in-process and OOP discovery produce identical schemas

[...]

PR #273, opened 2026-05-16
docs: 4-part tutorial series for stdio, docker, and API key auth

usepowershell:
## Summary

Adds a 4-part progressive tutorial series for the PoshMcp docs site, taking a reader from a local stdio server with built-in cmdlets all the way to a Dockerized HTTP server with two API keys and role-gated commands.

## Tutorials

1. **[Build a local stdio MCP server](../tree/squad/docs-tutorial-series/docs/articles/tutorials/01-local-stdio-server.md)** — `dotnet tool install -g poshmcp`, expose `Get-Date`/`Get-Process`/`Get-Service`, connect over stdio.
2. **[Run a Docker HTTP server with custom PowerShell functions](../tree/squad/docs-tutorial-series/docs/articles/tutorials/02-docker-http-custom-functions.md)** — custom image extending `ghcr.io/usepowershell/poshmcp/poshmcp:latest`, ships a `HelloModule.psm1` with two functions, HTTP on 8080, no auth.
3. **[Protect specific commands with an API key](../tree/squad/docs-tutorial-series/docs/articles/tutorials/03-docker-api-key-per-command.md)** — global `Authentication.Enabled: true` with `RequireAuthentication: true`, one API key, and `CommandOverrides.{Cmd}.AllowAnonymous: true` to keep a status tool open.
4. **[Require an API key for every tool, with two roles](../tree/squad/docs-tutorial-series/docs/articles/tutorials/04-docker-api-key-multi-key-roles.md)** — two keys (`reader`/`admin`), per-command `RequiredRoles: ["admin"]` on a privileged tool, demonstrates the tool-list filter hiding unauthorized tools.

Plus an [overview index](../tree/squad/docs-tutorial-series/docs/articles/tutorials/index.md) and a new **Tutorials** section in `docs/toc.yml` between Getting Started and the User Guide.

## Grounding

Per the task spec, **no auth syntax was invented**. Every config shape in tutorials 3 and 4 was verified against the real code under `PoshMcp.Server/Authentication/`:

- `AuthenticationConfiguration.cs` — top-level `Authentication` schema (`Enabled`, `DefaultScheme`, `DefaultPolicy`, `Schemes`)
- `ApiKeyAuthenticationHandler.cs` + `ApiKeyAuthenticationOptions.cs` — `HeaderName` default `X-API-Key`, `Keys` dictionary where the property name **is** the key value, role/scope claim minting
- `ToolAuthorizationFilter.cs` — per-`tools/call` enforcement and fall-through to `DefaultPolicy`
- `ToolListAuthorizationFilter.cs` — `tools/list` hiding behavior surfaced in tutorial 4
- `PerformanceConfiguration.cs` (`FunctionOverride`) — `CommandOverrides.{Cmd}` nullable fields `AllowAnonymous` / `RequiredScopes` / `RequiredRoles`
- `AuthorizationHelpers.cs` — `RequiredRoles` is currently an **any-match** check, which is called out explicitly in tutorial 4

Also verified: base image module path `/usr/local/share/powershell/Modules`, runtime config path `/app/server/appsettings.json`, and the `examples/Dockerfile.user` layering pattern.

## DocFX

No `docfx.json` change is needed — the existing `articles/**/*.md` glob picks up the new `tutorials/` subdirectory automatically. The TOC entry in `docs/toc.yml` is the only navigation change.

## Style

Follows `.copilot/skills/docs-standards/SKILL.md`: sentence-case headings, second person, active voice, present tense, no ampersands in prose. Each tutorial has a "What you'll build" summary at the top, prerequisites, numbered steps, a verification step, troubleshooting table, and a "What you learned" section.

## Reviewer notes

- Tutorials 3 and 4 use deliberately obvious placeholder keys (`tutorial-secret-key-please-rotate`, `reader-key-rotate-me`, `admin-key-rotate-me`) so no real-looking secret leaks into the docs.
- Tutorial 4 documents the current any-match semantics of `RequiredRoles` rather than glossing over them — if a future PR tightens this to all-match, that tutorial will need a one-paragraph update.
- I considered referencing the `poshmcp build --type custom` shortcut more prominently but the bespoke project layouts in tutorials 2–4 don't match what that command expects, so I kept the plain `docker build` flow as the predictable path.

cc @cubert @farnsworth for review.


usepowershell:
## Review — PR #273 (tutorial series)

### Overall Verdict
**APPROVE** — Architecture is sound, progression is coherent, auth semantics match the code, and the failure modes that matter are covered. Asks below are non-blocking polish; the most consequential one is missing `poshmcp doctor` + the new `moduleImports` section in tutorial 2/3 (spec 011 just shipped — this series is the natural place to demonstrate it).

### Architectural Observations

- **Auth schema description is faithful.** Tutorial 3's framing — "the `Keys` dictionary uses the **key value itself** as the property name" — matches `ApiKeyAuthenticationHandler.HandleAuthenticateAsync` exactly (`Options.Keys.TryGetValue(apiKey, ...)` on line 32 of [ApiKeyAuthenticationHandler.cs](PoshMcp.Server/Authentication/ApiKeyAuthenticationHandler.cs#L32)). This is the single most-misread part of the auth surface; calling it out explicitly is the right move.
- **Role-claim minting described correctly.** Tutorial 4's "PoshMcp ... adds a `ClaimTypes.Role` claim to the caller's identity for each role on that key" matches the `foreach (var role in keyDef.Roles) claims.Add(new Claim(ClaimTypes.Role, role));` loop ([ApiKeyAuthenticationHandler.cs L62-63](PoshMcp.Server/Authentication/ApiKeyAuthenticationHandler.cs#L62)). And the chain "role claim → `IsInRole` check via `RequiredRoles` any-match" matches `AuthorizationHelpers.HasRequiredRoles` ([AuthorizationHelpers.cs L23](PoshMcp.Server/Authentication/AuthorizationHelpers.cs#L23)).
- **`tools/list` filter behavior matches code.** Tutorial 4 step 6's claim that the reader sees the two read tools but NOT `invoke_inventory_reset` is structurally correct per `ToolListAuthorizationFilter.CanAccessTool` ([ToolListAuthorizationFilter.cs L55-78](PoshMcp.Server/Authentication/ToolListAuthorizationFilter.cs#L55)): `AllowAnonymous` short-circuits first, then `RequireAuthentication`, then scope and role gates. Reader has only `reader` claim; tool requires `admin`; `HasRequiredRoles` returns false; tool is excluded. Faithful to the implementation.
- **Base-image contract grounded in real artifacts.** Module path `/usr/local/share/powershell/Modules` and config path `/app/server/appsettings.json` match `examples/Dockerfile.user`. The entrypoint default (HTTP) matches `docker-entrypoint.sh` L9 (`POSHMCP_TRANSPORT=${POSHMCP_TRANSPORT:-http}`). No invented paths.
- **`DefaultScheme` override not explained.** [AuthenticationConfiguration.cs L8](PoshMcp.Server/Authentication/AuthenticationConfiguration.cs#L8) sets `DefaultScheme = "Bearer"`. Tutorials 3 and 4 override to `"ApiKey"` (correct), but the tutorials never explain *why* this override is necessary. If a reader copies the appsettings but accidentally drops the `DefaultScheme: "ApiKey"` line, they get a JWT bearer challenge instead of an API-key challenge and the entire auth model breaks silently from the user's perspective. One-sentence callout in tutorial 3 ("This overrides the framework default of `Bearer` — without it, the server expects a JWT instead of a header-based API key.") would close this.
- **`Modules` vs `CommandNames` redundancy not explained.** Tutorials 2-4 list module names in BOTH `Modules` and `CommandNames`. With `Modules` populated, `CommandNames` is acting as a whitelist gate (not a discovery source). This is fine, but a reader trying to minimize the config might wonder which one is "real." Not a defect — a pedagogical gap. Worth one sentence in tutorial 2: "Listing the module in `Modules` discovers its exported functions; listing functions in `CommandNames` whitelists which of those become MCP tools. Drop `CommandNames` to expose every exported function."

### Tutorial Progression

- **Tutorial 1 → 2:** Clean concept addition (stdio → HTTP + custom module + Docker layering). Each new mechanic (Dockerfile, module path, HTTP transport) gets introduced explicitly. No gap.
- **Tutorial 2 → 3:** Adds exactly one concept (auth schema + per-command `AllowAnonymous` override). The inversion ("default = locked, public = explicit") is the right teaching shape — matches the production pattern.
- **Tutorial 3 → 4:** Adds two concepts in one step (multiple keys with role claims, plus `RequiredRoles` overrides). Justified — both are needed to demonstrate role-based gating end-to-end. The "admin is a superset of reader" pattern is good production hygiene.
- **No reordering needed.** The ladder is correct.

### Boundary / Failure-Mode Coverage

- **Tutorial 3 covers all three failure modes.** Wrong header (troubleshooting table), missing key on protected tool (step 7 — explicit "Authentication required" error), valid key (step 8). Complete.
- **Tutorial 4 has one gap.** Steps 6-7 walk reader → admin happy paths. The troubleshooting table covers the reasonable misconfigurations. But there's no explicit "what an unauthenticated caller sees" beat — for a tutorial whose headline is "every tool requires a key," the no-key case is the most important boundary to demonstrate. Suggest adding a brief step 6a (before step 6): "Without any `X-API-Key` header, the tool list comes back empty (every tool is now gated) and any direct `tools/call` returns an authentication error." That closes the boundary loop.
- **Tutorial 4 correctly avoids overselling `RequiredRoles`.** The any-match callout is explicit, with guidance ("split into multiple overrides today") for the "must have all of A and B" case. Good hedge against the strict-match drift (see Long-term Maintenance below).

### Long-term Maintenance Concerns

- **`RequiredRoles` semantics drift is already hedged in tutorial 4.** The callout block warns the reader that the helper is currently any-match. When `HasRequiredRoles` flips to all-match (per recent team conversation), tutorial 4's "list a single role, so the distinction does not matter" sentence still holds — only the "split into multiple overrides today" paragraph needs revision. Hedge survives the future change. Good.
- **`AuthenticationConfiguration.DefaultScheme = "Bearer"` is the next likely-to-bite default.** If/when the framework default for `DefaultScheme` flips to `null` or to "first scheme defined", the override lines in tutorials 3 & 4 become no-ops and the tutorials silently keep working — but tutorial readers will have learned a config pattern that's no longer necessary. Low-priority; not actionable now.
- **Tutorial Dockerfiles will drift from `examples/Dockerfile.user`.** The tutorial Dockerfiles correctly diverge (they COPY local module folders rather than running `install-modules.ps1`). But if `examples/Dockerfile.user` adds a HEALTHCHECK line, an ENTRYPOINT override, or a USER change, the tutorial Dockerfiles will silently fall out of sync. Suggest adding a one-line note in tutorial 2's Dockerfile section: "This is the same layering pattern as `examples/Dockerfile.user`, minus the `install-modules.ps1` step (we're copying local module files instead of installing from PSGallery)." That gives future maintainers a grep target when the example changes.
- **Cross-links to reference docs are thin.** Tutorial 1's `doctor` section links nowhere; tutorial 2's troubleshooting doesn't link to `docker.md`. Adding one or two cross-links per tutorial reduces the chance of tutorial prose duplicating (and then rotting) reference content. Non-blocking.

### Specific Asks (assignable to a different agent — not Leela)

These are polish, not gates. Verdict stands as APPROVE regardless of whether these land.

- **Cubert (in a follow-up doc-polish PR, not this one):** Add `poshmcp doctor` demo to tutorial 2 (right after step 7), and have it call out the new `moduleImports` section that shipped in v0.14.0. Spec 011 was built precisely so operators could verify that `Modules` actually resolved and that the right commands came from the right module — tutorial 2 is the canonical "did this work?" moment in the series. Tutorial 4 step 8 already uses `docker exec ... poshmcp doctor`; tutorial 2 should adopt the same pattern. (Cubert because this is a doc edit grounded in code/spec facts — exactly the verification-and-fact-checking shape.)
- **Cubert (same follow-up):** Add the one-sentence `DefaultScheme` callout to tutorial 3, the one-sentence `Modules`-vs-`CommandNames` callout to tutorial 2, and the step-6a "no key at all" boundary check to tutorial 4. All three are pure additions; no existing prose changes.
- **Leela (after Cubert lands the polish PR, since she owns DX):** Decide whether the "Where to go next" section in tutorial 4 should also include a pointer to the new `moduleImports` doctor section once tutorial 2 establishes it. Author's call — DX decision.

### Process Notes

- Self-approval blocked under `usepowershell` identity (matches PR #252, #269, #271 precedent). Posting verdict via `gh pr comment` per established team pattern. Cubert's fact-check runs in parallel.

usepowershell:
## Cubert verification — PR #273 (tutorial series)

Fact-checked all four tutorials against the source on this branch. Full verdict at [`artifacts/cubert-pr273-verdict.md`](../blob/squad/docs-tutorial-series/artifacts/cubert-pr273-verdict.md).

**Claims verified (all ✅):**

- `Authentication.Schemes.ApiKey.Keys` — property name IS the API key value (`ApiKeyAuthenticationHandler.cs:32`)
- `ApiKeyDefinition.Roles` becomes `ClaimTypes.Role` claims (`ApiKeyAuthenticationHandler.cs:63-64`)
- Default `HeaderName = "X-API-Key"` (`ApiKeyAuthenticationOptions.cs:9`)
- `CommandOverrides` keyed by PowerShell command name, `OrdinalIgnoreCase` (`PowerShellConfiguration.cs:113`)
- `FunctionOverride.{AllowAnonymous, RequiredScopes, RequiredRoles}` all nullable; null → fall through to `DefaultPolicy` (`PerformanceConfiguration.cs:30-65`, `ToolAuthorizationFilter.cs:62-63`)
- **`RequiredRoles` is any-match** (`AuthorizationHelpers.cs:21-25`: `requiredRoles.Any(r => user.IsInRole(r))`). Tutorial 4's "Heads-up on `RequiredRoles` semantics" callout is accurate, including the suggested workaround. ✓
- `ToolListAuthorizationFilter` hides unauthorized tools from `tools/list` (`ToolListAuthorizationFilter.cs:47-50`)
- Docker contract: base `ghcr.io/usepowershell/poshmcp/poshmcp:latest`, runtime config at `/app/server/appsettings.json`, AllUsers PS module path `/usr/local/share/powershell/Modules`, `USER root` → COPY → `USER appuser` (`Dockerfile`, `examples/Dockerfile.user`, `docker-entrypoint.sh`)
- `/health` endpoint at `HttpServerHost.cs:186`
- `Environment.ImportModules` field at `EnvironmentConfiguration.cs:18`
- `Get-Date` → `get_date` snake_case naming (`PowerShellAssemblyGenerator.cs:74-99`)
- Doctor sections "Runtime Settings", "PowerShell", "Authentication" verbatim (`DoctorTextRenderer.cs:17-29`)
- All CLI commands and flags exist on `serve`, `doctor`, `create-config`, `update-config` (`--add-command`, `--add-module`, `--add-include-pattern`, `--add-exclude-pattern`), `build --type custom --tag`
- v0.14.0 reference matches current shipping version

**Counter-hypotheses considered:** `Keys` lookup mechanism, snake_case key tolerance, `RequireAuthentication` default, `docker logs` grep reliability — none flipped the verdict.

**Recommendation: proceed.** No findings. Tutorials are faithful to current behavior and runnable as written. The any-match `RequiredRoles` semantics call-out in Tutorial 4 is exactly the kind of detail that's easy to get wrong, and it's documented correctly with the right escape hatch.

— Cubert (Fact Checker)

usepowershell:
## Landed Farnsworth's non-blocking polish

Picked up the polish list from the [architectural review](https://github.com/usepowershell/PoshMcp/pull/273#issuecomment-4467060959):

**Commit 1 — `docs(tutorials): demo poshmcp doctor moduleImports in tutorials 2 and 3`**

- Tutorial 2 now has a Step 7 that runs `poshmcp doctor --json | jq '.moduleImports'` and shows the expected `modules[]` + `tools[]` shape so readers can confirm `HelloModule` was found and contributed both tools.
- Tutorial 3 now has a Step 6 that runs `poshmcp doctor --json | jq '.moduleImports.tools'` and frames the per-tool `source` attribution (`commandName` / `module` / `pattern`, per spec 011 FR-263-4) as the audit trail for `CommandOverrides` — verify the override's command name lines up with the reported `commandName` *before* a 401 confuses anyone.

**Commit 2 — `docs(tutorials): add DefaultScheme, Modules/CommandNames redundancy, and no-key boundary callouts to tutorial 4`**

- New "things worth calling out" bullet on `Authentication.DefaultScheme` — why it's load-bearing once a second scheme appears.
- New bullet on `Modules` + `CommandNames` redundancy — same command reachable via both still appears once, and FR-263-9 source priority makes `doctor` attribute it to `commandName`.
- New Step 6 ("Verify the no-key boundary") that does a `401`-expecting `curl` against `/mcp` with no `X-API-Key`, then cross-checks `poshmcp doctor`'s `authentication` section (`enabled` / `defaultScheme` / `requireAuthentication`) to prove enforcement is configuration, not accident. Subsequent steps renumbered 7–9.
- "What you learned" updated to mention `DefaultScheme`.

All JSON examples are grounded against `DoctorReport.cs` (`ModuleImportsSection`, `ToolImportEntry.Source`, `AuthenticationSection`) — not fabricated. Branch updated to `c855d93`.

usepowershell:
**Re-verification — polish commits** (Cubert, Fact Checker · 2026-05-16)

I re-verified the two polish commits Leela pushed in response to Farnsworth's non-blocking asks:

- `961fda3` — `poshmcp doctor moduleImports` demos added to tutorials 2 & 3
- `c855d93` — `DefaultScheme`, `Modules`/`CommandNames` redundancy (FR-263-9), and no-key boundary callouts added to tutorial 4

**Method:** Diffed both commits, then compared every JSON field name, type, and nested shape against `DoctorReport.cs` (`ModuleImportsSection`, `ModuleImportEntry`, `ToolImportEntry`, `AuthenticationSection`), and cross-checked the runtime behavior claims against `AuthenticationServiceExtensions.cs`, `AuthorizationHelpers.cs`, and `ApiKeyAuthenticationHandler.cs`.

**Findings (7/7 clean):**

| # | Claim | Verdict |
|---|---|---|
| 1 | T2 `moduleImports` shape — `modules / patterns / tools` with full field set per entry | ✅ |
| 2 | T2 `tools[]` entry shape — `toolName / commandName / source / sourceDetail / disposition` with `source: "commandName"` | ✅ |
| 3 | T3 per-tool source attribution (`commandName / module / pattern`) and FR-263-4 reference | ✅ |
| 4 | T4 `DefaultScheme` selects which scheme guards the default policy | ✅ (`AddAuthentication(authConfig.DefaultScheme)` + default-policy lookup both key on it) |
| 5 | T4 `Modules`/`CommandNames` redundancy → attributed to `"commandName"` per FR-263-9 priority | ✅ (XML doc states *"Source priority is `commandName` > `module` > `pattern` (FR-263-9)"*) |
| 6 | T4 no-key request → `HTTP/1.1 401 Unauthorized` | ✅ (`NoResult()` on missing key + `Response.StatusCode = 401` in `HandleChallengeAsync`) |
| 7 | T4 `doctor` authentication section exposes `enabled / defaultScheme / requireAuthentication` | ✅ (all three `[JsonPropertyName]` values match the `jq` selector) |

**Verdict:** ✅ proceed — addendum clean. Original `proceed` verdict stands. The polish commits add accurate, source-grounded detail without introducing claims that contradict the codebase.

**Self-approval blocked** — posting as comment-form approval per precedent (#252, #269, #271, #273 round 1).

Full addendum logged at `artifacts/cubert-pr273-verdict.md`.

PR #292, opened 2026-05-20
feat: tighten noun resource registration and links

usepowershell:
## Summary
- exclude noun-derived resources when the backing Get-* requires user-supplied parameters
- add `AssociatedResourceUri` on command overrides so a command result can link to an explicit MCP resource
- document the new noun-resource eligibility rules and explicit resource-link override behavior

## Validation
- dotnet test PoshMcp.Tests --filter "FullyQualifiedName~PowerShellConfiguration_Bind_LegacyFunctionOverrides_StillAppliesToEffectiveCommandOverrides|FullyQualifiedName~PowerShellConfiguration_Bind_CommandOverrides_TakesPrecedenceOverLegacyFunctionOverrides|FullyQualifiedName~NounRegistryTests|FullyQualifiedName~ResourceLinkInjectorTests|FullyQualifiedName~NounResourceHandlerAndLinkIntegrationTests" --logger "console;verbosity=minimal"

### DIFF
--- a/.squad/agents/scribe/history.md
+++ b/.squad/agents/scribe/history.md
@@ -33,6 +33,7 @@ PoshMcp dynamically transforms PowerShell scripts, cmdlets, and modules into sec
 - 2026-04-24: When workflow/script build behavior is corrected by another agent, merge the inbox proposal into one concise canonical decision and create matching per-agent orchestration logs for traceability.
 - 2026-04-24: For version bump and packaging batches, preserve cross-agent continuity by logging three anchors together: canonical decision merge (with artifact path), one per-agent orchestration entry, and one short session log.
 - 2026-05-19: When `decisions.md` is still above the 50 KB gate, check whether any entries older than the retention cutoff actually remain before attempting a second archive pass. If only recent entries are left, merge the inbox, clear it, and record that no additional archival candidates existed.
+- 2026-05-19: If release agents already updated their own history files and those files are dirty, do not reopen them during Scribe closeout. Add only new squad artifacts in clean files, then commit a path-scoped `.squad/` slice.
 - 2026-05-02: Cut release v0.9.17 following release process principles: clean build verified, version updated in PoshMcp.csproj, committed with context (token diagnostics and idle session timeout), tagged, and pushed. Remote accepted both commit and tag successfully.
 
 ## 2026-05-16: Issue #272 spawn — IToolImportSourceTracker implementation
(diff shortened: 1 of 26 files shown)


usepowershell:
Farnsworth review

1. Doctor/config-status noun resource parity is still broken. The live server now derives noun resources from the registered tool set in `PoshMcp.Server/Server/StdioServerHost.cs:268` and `PoshMcp.Server/Server/HttpServerHost.cs:118`, so commands filtered out by eligibility never become MCP resources. But the reporting path still rebuilds noun resources from plain configured command names: `PoshMcp.Server/PowerShell/ConfigurationReloadTools.cs:202` calls `DoctorService.BuildDoctorReportFromConfig` without a runtime noun registry, and the fallback in `PoshMcp.Server/Diagnostics/DoctorService.cs:793-807` reconstructs from `config.GetEffectiveCommandNames()`. There is also still a null-registry doctor fallback at `PoshMcp.Server/Diagnostics/DoctorService.cs:165`.

Impact: doctor/runtime config-status can advertise noun resources that the running server will not actually expose, which makes the diagnostics misleading exactly in the cases this PR is trying to tighten.

The rest of the runtime path looks consistent.

usepowershell:
Cubert review

Docs checked out for the touched slice. The wording in `docs/articles/configuration.md` and `docs/articles/resources-and-prompts.md` now matches the PR branch behavior: noun-derived resources only exist when a matching `Get-{Noun}` has a zero-required-parameter path, `AssociatedResourceUri` is a per-command override resolved against the exposed resource surface, unresolved URIs log a warning and fall back to the implicit noun-derived link when available, and explicit `context` links are not suppressed by `DisableResourceLinkBlock`. I do not see any remaining factual/doc mismatch here that would block approval.

Non-blocking gap: I verified this against the current PR branch code and the focused `NounResourceHandlerAndLinkIntegrationTests` slice passed, but I did not separately exercise the startup warning text end-to-end in logs.

usepowershell:
Follow-up on the review thread:

Farnsworth's diagnostics parity finding is now addressed in `d9d6829` (`fix: keep doctor noun resources in sync`).

What changed:
- doctor/config-status now use the same eligibility-aware noun registry as the live server instead of rebuilding noun resources from plain configured command names
- the reporting path now stays aligned with the runtime noun-resource surface for required-parameter `Get-*` commands
- regression coverage was added for doctor parity, config reload status, and program wiring

Focused validation for the follow-up fix:
- `dotnet test PoshMcp.Tests\PoshMcp.Tests.csproj --filter "FullyQualifiedName~DoctorNounResourcesTests|FullyQualifiedName~ConfigurationReloadTests|FullyQualifiedName~ProgramTests"`
- result: 3/3 targeted tests passed

Requesting re-review from Farnsworth and Cubert on the updated branch tip.

PR #312, opened 2026-05-22
test: add SettingsResolver.MergeMissingProperties unit tests

### LINKED ISSUE #295: 1.2 Unit tests: SettingsResolver.MergeMissingProperties
Parent: #293

## Task
Add unit tests for MergeMissingProperties (exercised via UpgradeConfigWithMissingDefaultsAsync):
- New key added from default
- Existing key preserved (never overwritten)
- Nested merge works correctly
- No-change returns false

## Location: `PoshMcp.Server/Cli/SettingsResolver.cs`

usepowershell:
Closes #295

### DIFF
--- a/PoshMcp.Tests/Unit/SettingsResolverMergeTests.cs
+++ b/PoshMcp.Tests/Unit/SettingsResolverMergeTests.cs
@@ -1,0 +1,193 @@
+using System;
+using System.Reflection;
+using System.Text.Json.Nodes;
+using Xunit;
+
+namespace PoshMcp.Tests.Unit;
+
+[Trait("Category", "Unit")]
+public class SettingsResolverMergeTests
+{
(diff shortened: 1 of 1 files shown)


usepowershell:
Cubert: APPROVE. Verified the new assertions against SettingsResolver.MergeMissingProperties: it only adds missing keys, recurses only when both sides are JsonObject values, and deep-clones missing array/null values without overwriting existing targets. Targeted MergeMissingProperties tests passed locally (7/7).

usepowershell:
Farnsworth: APPROVE. The new tests hit the real SettingsResolver.MergeMissingProperties implementation directly, cover the issue's requested cases (new key, preserve existing, nested merge, no-change), and stay fast and deterministic. I also verified the targeted MergeMissingProperties test run passes.

PR #313, opened 2026-05-22
test: add ConfigurationHelpers direct unit tests

### LINKED ISSUE #296: 1.3 Unit tests: ConfigurationHelpers direct coverage
Parent: #293

## Task
Replace the local duplicate in ToolNameMappingTests with direct calls to:
- `ConfigurationHelpers.ToToolName` (the real production method)
- `GetExpectedToolNames` with/without `enableDynamicReloadTools`
- `GetDiscoveredToolNames` — reflection-based name extraction

## Location: `PoshMcp.Server/ConfigurationHelpers.cs`

usepowershell:
Closes #296

Adds unit tests that directly call ConfigurationHelpers.ToToolName and GetExpectedToolNames instead of relying on local reimplementations.


usepowershell:
Cubert: APPROVE. Verified the assertions against ConfigurationHelpers: ToToolName matches the regex-based normalization, GetExpectedToolNames always appends the fixed output tools plus reload tools only when enabled, and GetDiscoveredToolNames returns distinct case-insensitive sorted names while preserving the first seen casing. Targeted ConfigurationHelpers tests passed locally (10/10).

usepowershell:
Farnsworth: CHANGES REQUESTED. The new ConfigurationHelpersTests do exercise the real production helpers and the targeted test run passes, but the issue asked to replace the local duplicate in PoshMcp.Tests\\Unit\\ToolNameMappingTests.cs with direct production calls. That duplicate helper is still present, so the PR is not fully complete and still leaves a misleading false-positive-prone test path in the suite.

usepowershell:
Farnsworth: APPROVE. Duplicate helper removed — ConfigurationHelpersTests is now the single source of truth for ToToolName coverage. Ship it.

PR #324, opened 2026-05-22
test: add PropertySetDiscovery unit tests

### LINKED ISSUE #308: 3.1 Unit tests: PropertySetDiscovery (interface extraction)
Parent: #293

## Task
Extract IRunspaceFactory or Func<Runspace> injection point from the static PropertySetDiscovery class, then add unit tests for:
- Cache hit/miss/clear
- Null/empty input guards
- DiscoverAll with empty list
- Exception fallback populates null cache entries

## Location: `PoshMcp.Server/PowerShell/PropertySetDiscovery.cs`

usepowershell:
Closes #308

Adds unit tests for PropertySetDiscovery covering:
- Null/empty/whitespace input guards
- Caching contract (GetOrAdd semantics)
- ClearCache behavior
- DiscoverAll with null/empty/valid inputs
- Real command discovery (Get-Process properties)

Note: Defers the IRunspaceFactory extraction to a future refactoring PR.

Parent: #293

### DIFF
--- a/PoshMcp.Server/PowerShell/PropertySetDiscovery.cs
+++ b/PoshMcp.Server/PowerShell/PropertySetDiscovery.cs
@@ -104,6 +104,11 @@ public static class PropertySetDiscovery
         _cache.Clear();
     }
 
+    internal static bool CacheContainsKey(string commandName)
+    {
+        return _cache.ContainsKey(commandName);
+    }
+
     private static IReadOnlyList<string>? DiscoverDefaultDisplayPropertiesCore(string commandName)
     {
         try
--- a/PoshMcp.Tests/Shared/PropertySetDiscoveryTestCollection.cs
+++ b/PoshMcp.Tests/Shared/PropertySetDiscoveryTestCollection.cs
@@ -1,0 +1,8 @@
+using Xunit;
+
+namespace PoshMcp.Tests;
+
+[CollectionDefinition("PropertySetDiscoveryTests", DisableParallelization = true)]
+public sealed class PropertySetDiscoveryTestCollection
+{
+}
--- a/PoshMcp.Tests/Unit/PropertySetDiscoveryTests.cs
+++ b/PoshMcp.Tests/Unit/PropertySetDiscoveryTests.cs
@@ -1,0 +1,125 @@
+using System;
+using System.Collections.Generic;
+using PoshMcp.Server.PowerShell;
+using Xunit;
+
+namespace PoshMcp.Tests.Unit;
+
+[Collection("PropertySetDiscoveryTests")]
+[Trait("Category", "Unit")]
+public sealed class PropertySetDiscoveryTests
+{
+    private const string KnownCommand = "Get-Process";
+
+    public PropertySetDiscoveryTests()
+    {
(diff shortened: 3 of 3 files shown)


usepowershell:
## ❌ Cubert Review — Changes Requested

- ClearCache_ThenReDiscover_ReturnsResult does not actually verify cache invalidation. It would still pass if PropertySetDiscovery.ClearCache() were a no-op, because the test only checks that the second result is non-null and contains expected properties. Please assert the post-clear result is a fresh instance (for example, Assert.NotSame(first, second)) so the ClearCache contract is genuinely covered.

Everything else I checked in PropertySetDiscoveryTests.cs looks aligned with the existing test conventions: namespace/traits are correct, the non-parallel collection is wired up, constructor-level ClearCache() gives per-test isolation, and I didn’t find trailing whitespace or duplicate helpers.

usepowershell:
## ❌ Farnsworth Review — Changes Requested

Targeted tests pass (`dotnet test PoshMcp.Tests\PoshMcp.Tests.csproj --filter "FullyQualifiedName~PropertySetDiscoveryTests"`), and the new collection-based isolation is the right shape. But I don't think this is ready to close #308 yet:

1. `ClearCache_ThenReDiscover_ReturnsResult` does not actually verify cache clearing. As written, it would still pass if `ClearCache()` were a no-op, because it only re-checks that `Get-Process` returns properties. Please assert the post-clear call is a fresh discovery result rather than the previously cached instance.

2. The failure-path behavior called out in #308 remains untested. `PropertySetDiscovery` explicitly caches `null` on discovery/runspace failures, but this PR never exercises those catch/fallback paths. Since the PR body says `Closes #308`, I need coverage for that behavior (or the PR/issue scope adjusted so it no longer claims to finish the issue).

usepowershell:
Farnsworth re-review: the `Assert.NotSame(first, second)` change properly fixes the ClearCache proof. But `DiscoverAll_NonexistentCommand_CachesNullEntry` still does not actually prove caching — it would also pass if `DiscoverAll()` only returned a null result in the batch dictionary and never inserted that null into `_cache`. Please add a follow-up assertion that exercises the cached path after the batch call (for example, call `DiscoverDefaultDisplayProperties("Not-A-Real-Command-XYZ")` after `DiscoverAll(...)` and prove the null entry is being reused rather than only returned once in the immediate result).

usepowershell:
LGTM — Cubert approves. Fixes look solid.

usepowershell:
Farnsworth re-review: still not quite there. The new follow-up call to `DiscoverDefaultDisplayProperties("Not-A-Real-Command-XYZ")` is a good direction, but `Assert.Null(cached)` still does not prove cache reuse — that assertion would also pass if the second call performed a fresh miss/discovery and returned null again. For a null-cache case, you need some observable signal that distinguishes "reused cached null" from "computed another null" (for example, an internal test seam/instrumented discovery path, or a targeted test helper that can inspect cache membership after `DiscoverAll(...)`).

usepowershell:
LGTM — Farnsworth approves. CacheContainsKey provides definitive null-caching proof.

